PR #3, opened 2017-03-17
linter: Adding missing ClojureScript this-as macro

djwhitt:
Adds a linter exception for this-as, a ClojureScript macro that lets you bind JavaScript 'this' to a local variable.


candid82:
Thanks!

PR #13, opened 2017-05-23
Allows for ignoring unused namespace warning

joelash:
New option in `.joker` to hide a warning on an unused namespace


candid82:
Cool! This will be useful for cases when namespaces are required for their side effects. I might make a few tweaks but in general lgtm.

Thank you!

joelash:
Perfect, thanks!

PR #14, opened 2017-05-26
Link to syntastic-joker vim plugin

manicolosi:


candid82:
This is great! Thank you!

PR #31, opened 2017-11-15
fix linter description in README

nicwest:
couple of typo's in the linter output format description in the README


candid82:
Thanks!

Issue #32, opened 2017-11-21
limitations of only being able to pass a single flag to the linter


nicwest:
I recently wrote a linter for https://github.com/w0rp/ale (a vim plugin for async linting). Joker works really really well for this due to it's fast startup speed and nice outputs.

However having spent a couple of days using it I started running into an issue where joker didn't seem to detect my `.joker` file. Looking deeper into the issue, I found it was because ale was writing the contents of my file to a `/tmp/xzy.clj` file then linting that, loosing the context of the original sources directory and thus its `.joker` file. 

Swapping the ale linter to pipe the files contents to stdin worked... until I changed directory and again lost the context of my working directory. finally frustrated I swapped the linter to just read the file off the disk and now it only works on saving the file.

Ideally I would like to add a flag to the linter to parse in the config directly:

`joker --lint --config='{:known-macros [bar/def-something foo/another-macro ...]}' <filename/stdin>`

or

`joker --lint --config='/path/to/file/.joker' <filename/stdin>`

or 

`joker --lint --working-dir='/path/to/file/' <filename/stdin>`

however [the current main](https://github.com/candid82/joker/blob/45137442e3125b9441cba00c91afb15c1f1e969a/main.go#L200-L229) doesn't leave much leeway in terms of adding additional flags.

Instead of parsing `os.Args` manually something along the lines of the [flag](https://golang.org/pkg/flag/) lib would give more flexibility. I had a mess about last night converting it use [pflag](https://github.com/spf13/pflag) just to see if it would work and it appears functional ([if a bit ugly](https://github.com/candid82/joker/compare/master...nicwest:flags)).

Is this something that seems reasonable? I don't want to break anything in the repl/interpreter.

candid82:
Thanks for working on this! The code looks fine to me, feel free to submit a PR. As for new flag, I think `--working-dir` for the case of reading from `stdin` makes the most sense and should solve the original problem. Also, feel free to submit a PR to add a link to your ale linter.

Thanks!

candid82:
This has been addressed by #34. Thanks @nicwest!

candid82:
Looks like this broke one important use case: when Joker script accepts its own flags. For example, if `script.joke` takes `-s` flat, running `joker script.joke -s` will result in an error: `unknown shorthand flag: 's' in -s`.

nicwest:
bugger, I was worried I would overlook something

I've written some tests, but it looks like we aren't going to be able to use any of the flag varients for this :(

```clj
[...]

PR #34, opened 2017-11-24
Flags

nicwest:
ref #32 

This PR uses https://github.com/spf13/pflag for flag parsing and adds the `--dialect` and `--working-dir` flags for better config detection when reading from stdin


candid82:
Great work, thanks @nicwest!

PR #44, opened 2017-12-28
add link to vscode extension

martinklepsch:
🍻


candid82:
Thanks! I am thinking about switching to VSCode myself, so this is very helpful. I do recommend updating the readme :)

Issue #77, opened 2018-09-11
Change Meaning of '--' on Command Line


jcburley:
The canonical meaning is "no arguments to the original command follow, so everything after this is either a real filename despite beginning with '-' or an option to a nested command".

Joker interprets it as "read from stdin", but the canonical way to express that is `-` (i.e. a single, not doubled, dash).

jcburley:
(Volunteering.)

jcburley:
Then I suggest adding support for `-` (single hyphen) as an "alias" for stdin, encouraging those other projects to convert to use it, and when they've done that (and suitable time has elapsed), finishing support for this issue.

When it comes to conventions, there's usually a good reason for them. In this case, it's possible Joker might never need to offer proper `--` support, if its command line is assumed to never support any further Joker (versus target-script) options after the name of the target script; if `-` works (as stdin) for the name of the target script; if a target script name beginning with `-` is to never be supported, or else via a distinct option (such as `--read`); if `--eval _expr_` can appear only once and is treated as a target-script specification, so all the arguments that follow it belong to the script, etc....

With all those caveats, especially given the youth of the project (unless you think it has a short overall lifetime!), I strongly recommend moving fairly aggressively to use the canonical form.

After all, Makefiles having "significant" tabs has been a huge hassle for decades, but was considered for change early -- yet rejected because of its "installed base" (of a "dozen" users!!):

https://stackoverflow.com/questions/1755550/what-is-the-reasoning-behind-the-makefile-whitespace-syntax

jcburley:
On the other hand, if this is more of a "proof of concept" project (and I do wonder if the name "Joker" suggests that?), there's probably no reason for these particular changes (`--` versus `-`') to be made.

jcburley:
See also https://github.com/candid82/joker/issues/32#issuecomment-420405105, which points out the "canonical" definition of `--` in `pflags`, a package that the PR referenced (to clean up options handling in Joker) but could not be fully merged, at that time, because `--` already meant something else.

candid82:
Thank you for the detailed answer. I think your suggestion on multi-step migration (make `-` an alias for  `--` first, migrate the clients, change the meaning of `--`) is the way to go. 
Joker is currently just a hobby project, but people do use it (mostly as a Clojure linter), so I do want to make the transition as smooth as possible.

PR #79, opened 2018-09-11
Support '-' (single hyphen) as alias for stdin

jcburley:
The documentation could be expanded and clarified further, but this is a quick PR to add support so a release could be made and downstream clients encouraged to switch over to using `-`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,6 +36,8 @@ You can also [build](#building) Joker from the source code.
 
 `joker --lint <filename>` - lint a source file. See [Linter mode](#linter-mode) for more details.
 
+`joker -` - execute a script on standard input (os.Stdin). (Supersedes `--`, which will be changed to mean "end of Joker options" in the future.)
+
 ## Documentation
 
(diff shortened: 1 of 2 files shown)


jcburley:
(This is the first step in resolving https://github.com/candid82/joker/issues/77.)

jcburley:
Is there anyplace (else) where `--` is documented as meaning **stdin**? I haven't noticed any...maybe nobody's actually using it?

candid82:
It's mentioned [here](https://github.com/candid82/joker#linter-mode). It's used by https://github.com/candid82/flycheck-joker and https://github.com/candid82/SublimeLinter-contrib-joker and possible by other plugins (I don't own them, I'll need to check).

Thanks!

jcburley:
Ah, I missed that in the README.md, despite ooking twice! Fixing now.

jcburley:
A next step (before or after a release, including pinging known clients?) might be issuing a warning if `--` is used.

(Again, volunteering.)

PR #81, opened 2018-09-12
Warn if '--' is used for a filename

jcburley:
Per https://github.com/candid82/joker/pull/79#issuecomment-420516049.


candid82:
Thanks!

PR #87, opened 2018-09-14
Its 'fourth', not 'forth'

jcburley:
(As in the sequence"first, second, third, **fourth**, fifth, sixth, ...".)


candid82:
Thanks!

PR #88, opened 2018-09-14
Add CPU and Memory Profiling Support

jcburley:
Includes (optional) setting of pertinent rates.


jcburley:
Looks like `go get pkg/profile` needs to be added to the CI build setup.

candid82:
Yep, could you please update `.circleci/config.yml` accordingly?
Also, what are the differences between `pkg/profile` and `runtime/pprof`? Does `pkg/profile` bring enough value to justify an additional external dependency?

jcburley:
That's a good question -- was wondering that myself after submitting the PR! I had incorporated both into another project when I was learning about profiling, and while I've tested and tried out both, I'm not aware of any salient differences.

I think I'd rather simplify things by removing `pkg/profile` than add another dependency, but am open to just adding the dependency and saving the simplification for later, when I (or someone) have time -- as things are about to get very busy for me through the weekend at least.

candid82:
Let's merge this as is for now. I'll play with both profiling options when I have time and if `pkg/profile` doesn't add much value over `runtime/pprof` we can always remove it later.

candid82:
Thank you!

PR #90, opened 2018-09-14
Allow env var to specify Joker executable to run, and give more info

jcburley:
Also set nonzero exit code for `(wrong exit code)` case.


candid82:
I am not sure I understand the use case for `JOKER_EXE`. Don't we want to always use `joker` binary from `joker` project directory (which is supposedly the latest built version) to run tests against?

jcburley:
I like to do A/B testing in context of the environment when confused by test results or behavior. Now that I've got a better sense of what that script is doing, and understand build procedures better, it seems less necessary. Though I personally try to avoid baking assumptions (like where executables live) into script balls of mud, this one doesn't belong to that category.

Even outputting the result seems less important now that I've seen it a few times and figured out that the error output is already being shown.

Is it wrong to ensure a nonzero result code is returned for that (when...) block though, I wonder?



On September 15, 2018 11:48:17 AM EDT, Roman Bataev <notifications@github.com> wrote:
>I am not sure I understand the use case for `JOKER_EXE`. Don't we want
>to always use `joker` binary from `joker` project directory (which is
>supposedly the latest built version) to run tests against? 
>
>-- 
>https://github.com/candid82/joker/pull/90#issuecomment-421589692

-- 

candid82:
I think the addition of `(var-set #'exit-code 1)` line is a valid fix.

jcburley:
Took a few moments to clarify the diagnostics and remove the `JOKER_EXE` override (but still moving determining the executable name to outside the loop, for clarity more than anything).

candid82:
Thanks!

Issue #94, opened 2018-09-30
Build Broken (#2)


jcburley:
Getting this on Ubuntu and Mac OS X:

```
[1] craig@doe:~/.go/src/github.com/candid82/joker (master * u=)$ ./run.sh -e '(println "i am here")'
Generating core.joke
Generating time.joke
Generating math.joke
Generating linter_all.joke
Generating linter_cljx.joke
Generating linter_clj.joke
Generating linter_cljs.joke
std/http/http_native.go:111: declaration of "h" shadows declaration at std/http/http_native.go:109
[1] craig@doe:~/.go/src/github.com/candid82/joker (master * u=)$
```

PR #95, opened 2018-09-30
Fix build error due to shadowed variable

jcburley:
Fixes: https://github.com/candid82/joker/issues/94

### DIFF
--- a/std/http/http_native.go
+++ b/std/http/http_native.go
@@ -106,9 +106,9 @@ func mapToResp(response Map, w http.ResponseWriter) {
 	if ok, b := response.Get(MakeKeyword("body")); ok {
 		body = AssertString(b, "HTTP response body must be a string").S
 	}
-	if ok, h := response.Get(MakeKeyword("headers")); ok {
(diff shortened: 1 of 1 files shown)


candid82:
Hm... For some reason I am not getting this error. CI build is also passing... I wonder if we use different versions of `vet`. What's the output of `go tool vet -V` on your machine?

jcburley:
```
vet version go1.11
```

(I upgraded to Go 1.11 last night on my main Ubuntu development machine after installing it on my Ubuntu server; already had it on my OS X laptop.)

jcburley:
Also I later noted that some earlier code, in the same file, used `headers` in a similar situation where I'm using `hs`, so if you want to change that for consistency, that might make sense. (I could do it too, but am up against a hard deadline early tomorrow morning, resuming sketchy availability all this coming week.)

candid82:
I'll merge and change it to `headers` since it makes sense for consistency. Still not sure what's causing the discrepancy in `vet` behavior though...

Issue #99, opened 2018-10-10
Integer Exit Code Not Returned by joker.os/sh 


jcburley:
Clojure provides the actual exit code returned by the program run via:

```
user=> (require '(clojure.java [shell :as sh]))
nil
user=> (sh/sh "sh" "-c" "exit 5")
{:exit 5, :out "", :err ""}
user=> (sh/sh "sh" "-c" "exit 0")
{:exit 0, :out "", :err ""}
user=>
```

Whereas Joker returns it only embedded within an (undocumented) string:

```
user=> (require '(joker [os :as os]))
nil
user=> (os/sh "sh" "-c" "exit 5")
{:success false, :err-msg "exit status 5", :out "", :err ""}
user=>
```

While `:err-msg` is nice, it should be documented, and an `:exit` key/value should also be returned with the actual error code.

jcburley:
FYI, I've hit a major milestone here in that, with the PR (my branch), I'm now using a Joker script on my Internet-facing server!

The script is one small piece of my arcane, but generally effective, SMTP server, including spam filtering and the like -- nothing worth publishing at this time. For now, it's invoked along with the original `bash` script from which it was converted, and the outputs compared (this is probably overkill at this point).

But it's a start, and it did require me to increase memory limits (more specifically, I had to change the the **daemontools** `softlimit` command line to use `-d` and `-s` instead of just `-m`, for which no value seemed to support Joker running -- something weird about including `-a` with `-d` and/or `-s`, which `-m` does automatically, because they worked independently).

I plan to slowly "evolve" more of my own scripts over to Joker scripts, then probably more of the main mailserver code, and at some point in there start really leveraging the power of Clojure. (Some of the conversions might well be to ClojureScript or Clojure proper; I needed Joker to handle the performance requirements for oft-invoked scripts "on the edges" of the architecture.)

Thanks again for creating Joker!

candid82:
This is great! Glad Joker is being used for scripting, which is certainly one of the use cases I had in mind for it (although most people use it purely as Clojure linter). We use it for “Devops” scripting on our team at work. We are a Clojure centric team and being able to write scripts in (almost) Clojure is nice.

candid82:
Fixed by https://github.com/candid82/joker/pull/100. Thanks @jcburley!

PR #100, opened 2018-10-10
Also return integer exit code from joker.os/sh (and sh-from)

jcburley:
This fixes:

https://github.com/candid82/joker/issues/99

However, it also updates re-generated files, which I'm not sure is properly included in this PR.

Please advise.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ joker
 .vscode
 .idea
 core/a_*.go
+*.prof
--- a/.gitignore
+++ b/.gitignore
@@ -2,4 +2,3 @@ joker
 .vscode
 .idea
 core/a_*.go
-*.prof
--- a/docs/index.html
+++ b/docs/index.html
@@ -35,6 +35,9 @@
   <a href="#joker.time">joker.time</a>
 </li>
 <li>
+  <a href="#joker.url">joker.url</a>
+</li>
+<li>
   <a href="#joker.yaml">joker.yaml</a>
 </li>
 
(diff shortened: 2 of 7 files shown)


candid82:
Hey James, I am on vacation away from computers. I’ll review / merge this PR in about a week. Thank you for your contributions!

jcburley:
No rush -- enjoy your vacation!

candid82:
This looks great, thank you! I usually update (re-generate) the docs before releasing, but sometimes I forget. The only downside of re-generating the docs with the code changes is that [standard library reference](https://candid82.github.io/joker/) will reflect master rather than latest released version. Given current adoption (which I suspect is close to non-existent outside of linting) this should not be a problem at the moment.

jcburley:
Great -- I'll refrain from re-gen'ing the docs in the future. I looked for some kind of master script that ran all the re-gen scripts, but didn't notice one in the repo. Maybe one named `prepare-release` or something like that would be good, and/or have the auto-gen'ed stuff comment whether they're release-generated or build-time-generated? Not sure what is best here....

Meanwhile, my 2nd Joker script is now in production on my server as of several hours ago, and working quite well. Overall it's so much simpler than the Bash script(s) it's replacing...but still fairly simple stuff.

candid82:
There is a script for preparing releases called release.joke and I [just added](https://github.com/candid82/joker/commit/602fc0724b78268e640f723b9ab1f8496604ccc8) the code to regenerate docs to it. Thanks for the suggestion!

Issue #109, opened 2018-10-26
The -e Option Should (println ...) The Result


jcburley:
For consistency with Clojure and ClojureScript (well, `lumo` anyway), maybe the behavior should be changed to print the result, if not `nil`.

PR #111, opened 2018-10-26
Print result of evaluating -e <expr> if not nil

jcburley:
Fixes: https://github.com/candid82/joker/issues/109

Note that this leaves `main.go` in a bit of a dodgy state in terms of code cleanliness, because `EVAL` not only means different things (as it did before), but there's a shiny new `Phase` value, `PRINT_IF_NOT_NIL`, that has a new meaning, but in only one context.

I'm up for cleaning that up, but wanted to keep this PR scoped to just fixing the issue at hand.

### DIFF
--- a/core/procs.go
+++ b/core/procs.go
@@ -38,6 +38,7 @@ const (
 	READ Phase = iota
 	PARSE
 	EVAL
+	PRINT_IF_NOT_NIL
 )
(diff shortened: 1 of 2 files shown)


candid82:
I think this makes sense. Thanks!

Issue #110, opened 2018-10-26
Rename `--expr` to `--eval`


jcburley:
Though there's already an `--evaluate` option, which perhaps should also be renamed if we do this, it might be a good idea to rename the long form of `-e` to `--eval` (from `--expr`), for consistency with Clojure and Lumo.

candid82:
+1

PR #113, opened 2018-10-27
Rename the --expr option to --eval

jcburley:
This is for consistency with Clojure and Lumo.

Fixes: https://github.com/candid82/joker/issues/110


candid82:
Thanks!

PR #114, opened 2018-10-27
Add tests of the --eval flag

jcburley:
One is "fake" and wouldn't work anyway because of limitations in how
the `testing` function parses string args. I'm leaving it in for now
in case I get time to try adding support for a test passing a vector
of args instead of a string, or some other simple solution.


candid82:
Thanks!

PR #119, opened 2018-10-31
Change Build Instructions in README.md

jcburley:
Pass `-d` to `go get` to prevent premature building; use `run.sh` script instead, and if it works, complete the installation.


candid82:
Thanks!

PR #123, opened 2018-11-15
Change "--" to mean only end of Joker options, never stdin

jcburley:
Fixes https://github.com/candid82/joker/issues/77.

(Seems like enough time, and at least one release, have passed since this was deprecated.)


candid82:
Thanks!

PR #124, opened 2018-11-15
Primitive Socket Repl Support

jcburley:
Thought I'd try this out to work around delve (`dlv`) refusing to debug programs that read from `stdin`, and to learn more about Go and Joker internals.

This seems much more intrusive than necessary, but (try as I might -- and I spent way too much time on this!!) I can't figure out how to simply replace `os.Stdout` and `os.Stderr` with implementations that work over TCP.

But the result seems to work fairly well: I can start up a `dlv debug -- -nrepl localhost:2000` session in one window, start it running, then `telnet localhost 2000` in another window and interact with Joker from there (while doing `dlv` stuff in the first window).

Downsides include that there's no `readline` support using this approach, although `rlwrap telnet` works pretty well, though the only graceful way I can find to exit that is `(joker.os/exit 0)` (maybe a shortcut like `:repl/quit` should be supported?).

If this isn't suitable (even at the conceptual or design levels) for merging, I'll probably keep it around in my own fork to make things easier when I debug. Supposedly the "right way" to debug in `dlv` involves attaching the debugger to an existing process, but I've read claims that it doesn't always work, and I prefer my debugging setup to always work.


snoe:
Awesome. 

In case this was overlooked, however, I think this looks much more like clojure's socket repl than nrepl.

I believe nrepl messages usually take the form `{"op" "eval" "code" "(+ 1 2 3)"}` whereas the socket server reads sexpr strings (by default).

https://clojure.org/reference/repl_and_main#_launching_a_socket_server

jcburley:
> Awesome.
> 
> In case this was overlooked, however, I think this looks much more like clojure's socket repl than nrepl.
> 
> I believe nrepl messages usually take the form `{"op" "eval" "code" "(+ 1 2 3)"}` whereas the socket server reads sexpr strings (by default).
> 
> https://clojure.org/reference/repl_and_main#_launching_a_socket_server

Ah, okay, thanks for clearing that up! I'd seen that form but hadn't associated that with the nature itself of an nrepl.

I'll look into how socket repl's are handled and rework this PR to be more consistent with that terminology and approach.

(Probably I was thinking "nrepl's are cool; this new feature is cool; so this new feature must be an nrepl!  ;) )

jcburley:
I forgot to handle `GLOBAL_ENV.stdin.Value` too, so this doesn't work from the socket REPL:

```
$ rlwrap telnet localhost 2000
Trying 127.0.0.1...
Connected to localhost.
Escape character is '^]'.
Welcome to joker v0.10.1, client at 127.0.0.1:37598. Use '(joker.os/exit 0)', or close the connection, to exit.
user=> (read-line)
""
user=>
```

I'll try to make time to fix that later today, either in this PR or (if it has already been merged) in a separate PR.

jcburley:
`(read-line)` works now, in standalone mode, in the socket repl, and in the socket repl running under `dlv`.

I'm still kinda frustrated that I couldn't just stick replacement instances in `os.Std*` -- tried various approaches in Joker itself, then in a small sandbox program. I feel like I must be missing something obvious, but am getting no reassurance of that by checking on the interwebs, as nobody else seems to have come up with the kind of simple solution I was expecting....

(Go is wonderful and all, but sometimes I think it'd be better named Pause, or Stop!)

candid82:
I like the idea, but the implementation looks awkward to me. Specifically, `JokerWriter`, a wrapper around either `os.File` or `net.Conn`, seems unnecessary. As far as I can tell, we really only need  wrappers for`io.Writer` and `io.Reader`, which both `os.File` and `net.Conn` implement. Global vars could look like 

```
var (
	Stdin = io.Reader
	Stdout = io.Writer
	Stderr = io.Writer
)
```

`Env.stdin.Value`, `Env.stdout.Value` and `Env.stderr.Value` would hold aforementioned wrappers.
Perhaps it doesn't even make sense to create wrappers for each individual interface and instead it can be a generic wrapper called, say, `GoObject`, holding a value of type `interface{}`. The code then would do type assertions as necessary. 
I have not tried to implement this approach so it's possible I am missing something.

Thoughts?

jcburley:
I don't really understand what you're suggesting -- my Go skills are still rather poor, I'm afraid -- other than I agree the implementation was pretty bad, so I've tried various approaches in a sandbox and now have something that seems a bit better and still works, based on some quick testing.

Basically I can't figure out how to _not_ create a wrapper (now `io_writer.go`) to parallel `buffered_reader.go`, nor is it obvious to me how to make a single `GoObject` wrapper work (in the sandbox I create as a distinct, albeit tiny, repo, separate from this project).

(Note that I'll have limited availability for the next 8+ days.)

jcburley:
In the meantime, I much prefer the `Stdin`-versus-`JokerStdin` nomenclature you suggested, and this latest version (of the PR) gets rid of the thing I disliked the most about my original version, namely, the wrapper having both `os.File` and `net.Conn` members plus a runtime test as to which was populated.

It looks and feels cleaner to me now, and -- assuming we can trust Go to ensure that a `var` initializer that refers to another `var` will always run _after_ any initializer for that latter var (as is expected by the `parse.go` code that sets `GLOBAL_ENV` based on `NewEnv(...Stdin, Stdout, Stderr)`, which in turn depends on those three vars having already been set by their initializers in `environment.go`) -- I think it'll work well.

BTW that "playground" to which I referred, above, is here: https://github.com/jcburley/whatstd

core/environment.go
```diff
@@ -59,12 +67,12 @@ func NewEnv(currentNs Symbol, stdout *os.File, stdin *os.File, stderr *os.File)
 	res.CoreNamespace.meta = MakeMeta(nil, "Core library of Joker.", "1.0")
 	res.ns = res.CoreNamespace.Intern(MakeSymbol("*ns*"))
 	res.ns.Value = res.EnsureNamespace(currentNs)
-	res.stdout = res.CoreNamespace.Intern(MakeSymbol("*out*"))
-	res.stdout.Value = &File{stdout}
 	res.stdin = res.CoreNamespace.Intern(MakeSymbol("*in*"))
 	res.stdin.Value = &BufferedReader{bufio.NewReader(stdin)}
+	res.stdout = res.CoreNamespace.Intern(MakeSymbol("*out*"))
+	res.stdout.Value = &IOWriter{io.Writer(stdout)}
```
candid82: `stdout` is already of type `io.Writer`, no need to convert it to `io.Writer` again (the same goes for `stderr`, here and `SetStdIO`)

jcburley: Will check that out later, thanks!!

Issue #126, opened 2018-11-18
(read-line) Doesn't Read Entire Line


jcburley:
Clojure itself reads a complete line:

```
$ clojure
Clojure 1.9.0
user=> (read-line)
this is a test
"this is a test"
user=>
```

Joker not so much:

```
$ joker
Welcome to joker v0.10.0. Use EOF (Ctrl-D) or SIGINT (Ctrl-C) to exit.
user=> (read-line)
hey this is weird
"hey"
user=> (read-line)
"his"
user=> (read-line)
"s"
user=> (read-line)
"eird"
user=> (read-line)
what??
"what??"
user=> (read-line)
when 
"when"
user=> (read-line)

""
user=> (read-line)
""
user=> (read-line)
""
user=>
```

PR #127, opened 2018-11-18
Fix (read-line) to read entire lines

jcburley:
Fixes: https://github.com/candid82/joker/issues/126

### DIFF
--- a/core/data/core.joke
+++ b/core/data/core.joke
@@ -2149,7 +2149,7 @@
 
 (def
   ^{:arglists '([])
-    :doc "Reads the next line from *in*."
+    :doc "Reads the next line from *in*. Returns nil if an error (such as EOF) is detected."
     :added "1.0"}
   read-line read-line__)
 
--- a/core/procs.go
+++ b/core/procs.go
@@ -1103,11 +1103,24 @@ var procReadString Proc = func(args []Object) Object {
 	return readFromReader(strings.NewReader(EnsureString(args, 0).S))
 }
 
+func readLine(r *BufferedReader) (s string, err error) {
+	for {
+		line, isPrefix, err := r.ReadLine();
+		if err != nil {
+			return "", err
+		}
+		s += string(line)
+		if !isPrefix {
+			break
+		}
+	}
+	return
+}
+
 var procReadLine Proc = func(args []Object) Object {
 	CheckArity(args, 0, 0)
-	var line string
-	f := AssertIOReader(GLOBAL_ENV.stdin.Value, "")
-	fmt.Fscanln(f, &line)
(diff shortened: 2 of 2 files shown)


core/data/core.joke
```diff
@@ -2149,7 +2149,7 @@
 
 (def
   ^{:arglists '([])
-    :doc "Reads the next line from *in*."
+    :doc "Reads the next line from *in*. Returns nil if an error (such as EOF) is detected."
```
candid82: If I read the docs correctly, `ReadLine` won't return an error on EOF: `No indication or error is given if the input ends without a final line end.`. Which means this comment is not accurate: `read-line` will return the input until EOF, not `nil`, and I think this is how Clojure works as well.

core/procs.go
```diff
@@ -1103,11 +1103,27 @@ var procReadString Proc = func(args []Object) Object {
 	return readFromReader(strings.NewReader(EnsureString(args, 0).S))
 }
 
+func readLine(r *BufferedReader) (s string, err error) {
+	for {
+		line, isPrefix, err := r.ReadLine();
```
candid82: Would not using `ReadString('\n')` instead result in a simper implementation?

jcburley: Looks like it, thanks! I'll play around with that more when I have some free time, maybe later today. It'd be great to avoid adding my `readLine()` function, if it isn't needed.

jcburley: Upon further review, it seems `ReadString('\n')` is not equivalent, because `ReadLine()` accounts for a potential `\r` preceding the `\n`.

Would you prefer the current implementation (in this PR), or one that replaces my implementation of `readLine()` with one that calls `ReadString('\n')` and then removes any final `\n` and, if so removed, any resulting final `\r`, before returning the resulting string?

I'm not 100% sure the results would be exactly the same in all cases (the `bufio.ReadLine()` code does some non-trivial stuff), and in general I'd prefer to leave it to Go's runtime to decide what constitutes a "line" of text (while keeping the underlying buffer "in sync" so as to handle subsequent calls to `ReadLine()` or anything else that reads from it)...but I'm open to eliminating the current loop in favor of having Go's runtime loop for us and then just cleaning up afterwards.

(I guess we can be grateful we don't have to support systems using only `\r` to end lines -- or, for that matter, EBCDIC....)

candid82: I think current implementation is fine if the behavior is consistent with Clojure. Merging.

candid82:
Spoke too soon. This breaks this code `(with-in-str "sadfsaf" (read-line))`, because stdin is bound to `Buffer` inside `with-in-str`, not `BufferedReader`. Perhaps we need to use `bufio.Scanner` that works with any `io.Reader`...

Issue #129, opened 2018-11-21
(read-line) Fix Breaks (with-in-str ... (read-line))


jcburley:
E.g.:

```
$ ./joker
Welcome to joker v0.10.1. Use EOF (Ctrl-D) or SIGINT (Ctrl-C) to exit.
user=> (with-in-str "sadfsaf" (read-line))
panic: interface conversion: io.Reader is *core.Buffer, not *core.BufferedReader [recovered]
	panic: interface conversion: io.Reader is *core.Buffer, not *core.BufferedReader [recovered]
	panic: interface conversion: io.Reader is *core.Buffer, not *core.BufferedReader

goroutine 1 [running]:
main.processReplCommand.func1(0xc0006564a0)
	/Users/craig/.go/src/github.com/candid82/joker/main.go:106 +0x2e5
panic(0x13f0960, 0xc000730ed0)
[...]
```

jcburley:
Discovered per: https://github.com/candid82/joker/pull/127#issuecomment-440560002

PR #130, opened 2018-11-22
Fix (read-line) so it works inside (with-in-str ...)

jcburley:
This is not the elegant solution for which we were hoping, but it seems to work well, the key being to ensure the underlying `io.Reader` object, whatever its concrete implementation, is maintained such that a `(read-line)` reads only as far as necessary ("unreading" as appropriate) so the next `(read-line)` (or `(read)`) call works properly.

Fixes: https://github.com/candid82/joker/issues/129

### DIFF
--- a/core/procs.go
+++ b/core/procs.go
@@ -1127,6 +1127,8 @@ func readLine(inp io.Reader) (s string, e error) {
 				}
 			}
 			s = s[0:l]
+		} else if s != "" && e == io.EOF {
+			e = nil
 		}
 	default:
 		msg := fmt.Sprintf("Expected %s, got %T", "*core.BufferedReader or *core.Buffer", inp)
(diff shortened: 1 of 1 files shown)


core/procs.go
```diff
@@ -1103,23 +1103,43 @@ var procReadString Proc = func(args []Object) Object {
...
+			s += string(line)
+			if !isPrefix {
+				break
+			}
 		}
-		s += string(line)
-		if !isPrefix {
-			break
+	case *Buffer: // Underlying type is *bytes.Buffer
+		s, e = r.ReadString('\n')
```
candid82: Would not this code work for `bufio.Reader` as well (since it also has `ReadString` method)? If so, `readLine` could be simplified by accepting an interface with `ReadString` method...

candid82:
Thanks!

PR #134, opened 2018-12-03
Add module

bcarrell:
This adds Go module support (for Go >=1.11). 

This makes it cleaner to build `joker` in package managers because we can omit setting up `GOPATH` and can build directly in the directory.


candid82:
Thanks Brandon!

Issue #131, opened 2018-11-23
Create Test Suite for Evaluation of Expressions


jcburley:
E.g. test the various combinations of reading from files versus strings, `\n` versus `\r\n` line endings, multiple lines, lack of final newlines, etc., when using `(read-line)`, potentially in combination with `(read)` and such, per:

https://github.com/candid82/joker/issues/129

jcburley:
Could be kicked off by an `eval-tests.sh` script, to parallel `flag-tests.sh` and such.

Would add that to `.circleci/config.yml` as a final step.

candid82:
Thank you for working on this!

PR #136, opened 2018-12-04
New script to test evaluation of forms from input files

jcburley:
Pretty much just `linter-tests.sh` and friends, modified to support testing of fairly arbitrary input files (named `input.joke`) that are evaluated.

An `input.joke` file should be able to easily shell out and run `joker` on some other file in the same directory, as it is run in the test directory and can look at the first element of `(joker.os/args)` to find the name of the executable.

This is crude but promises to be effective as a start.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -34,3 +34,6 @@ jobs:
       - run:
           name: flag tests
           command: ./flag-tests.sh
+      - run:
+          name: eval tests
+          command: ./eval-tests.sh
--- a/eval-tests.sh
+++ b/eval-tests.sh
@@ -1,0 +1,3 @@
+#!/usr/bin/env bash
+
+./joker tests/run-eval-tests.joke "$@"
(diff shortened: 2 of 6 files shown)


jcburley:
Starting-point to address https://github.com/candid82/joker/issues/131.

jcburley:
Note that I've been investigating bringing `deftest` and friends from Clojure into Joker, because this PR is fairly heavyweight for a lot of useful tests that could just be copied over wholesale, in that the Joker executable being run could perform the tests directly instead of invoking itself (though that's a useful capability as well).

There's some work to do to support `deftest`, though. I'm not yet sure what, of its features, should be deferred versus implemented right away.

I think the first (and only) test added by this PR is an example of one that doesn't need a distinct Joker invocation nor separate files, so it would seem a fine candidate for `deftest`.

candid82:
Thanks! I agree porting `clojure.test` over to joker would be great.

PR #137, opened 2018-12-04
Improve std/generate-std.joke Script

jcburley:
Avoid conflicts between argument names and internally-generated (reserved) names by prefixing the reserved names with single underscores. (There might be a better approach to this that could be used; this assumes APIs won't use underscore-prefixed argument names., which might be an erroneous assumption.)

Support docstrings with embedded backticks. (Needed by `gostd2joker` work, but a reasonable thing to do now.)

Support multi-level namespaces, e.g. `joker.go.net`. (Needed by `gostd2joker` work, but a reasonable thing to do now.)

Optionally comment out Go code for "empty" packages. (Wanted for `gostd2joker` work in the past; not sure it'll be needed going forward.)

### DIFF
--- a/std/fn.tmpl
+++ b/std/fn.tmpl
@@ -1,9 +1,9 @@
-var {fnName} Proc = func(args []Object) Object {
-	c := len(args)
+var {fnName} Proc = func(_args []Object) Object {
+	_c := len(_args)
 	switch  {
 	{arities}
 	default:
-		PanicArity(c)
+		PanicArity(_c)
 	}
 	return NIL
 }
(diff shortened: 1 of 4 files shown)


jcburley:
Maybe I should bring some of my `run.sh` changes over from the `gostd2joker` branch (as a distinct PR)?

candid82:
Thanks! 
RE: `run.sh` changes. Looks like you call generate-docs and generate-std in your version. I was hesitant to do that because I thought it was not necessary most of the time and it would slow down the build, but perhaps the overhead is negligible (seems to be less than half a second). Feel free to send a PR.

PR #138, opened 2018-12-08
Fix e.g. (use 'joker.math)

jcburley:
`joker.math/pi` works in the current version, but this fix allows one to `(require '[joker.math :as m])` and such. Ditto the other libs added.

(Also sorted the libs so it's easier to see what's missing.)


candid82:
Thanks!

Issue #105, opened 2018-10-13
Support Class Paths


jcburley:
`(require ...)`, `(load ...)`, et al seem to load solely from the current directory (CWD), rather than searching through any sort of class path.

For production use, being able to put common `*.joke` files in a central library, and load them from there based on an environment variable (such as `JOKER_CLASSPATH`) containing a `$PATH`-style list of directories (semicolon-delimited on Windows, colon-delimited elsewhere), would allow Joker code to avoid having to hardcode the path(s) of such libraries.

candid82:
Implemented by https://github.com/candid82/joker/pull/152. Thanks!

PR #139, opened 2018-12-09
Support Class Paths

jcburley:
This addresses https://github.com/candid82/joker/issues/105, and also adds a barebones `joker.repl` (`doc` works).

There's a lot to unpack here, including various random decisions I made about how things should work, while trying to be compatible with existing installations and usage of Joker.

As I couldn't get a handle on how the Go ecosystem handles installation of artifacts that accompany (but are not to be embedded in) the executable, I kinda winged it, in part by creating an `install.sh` script that wraps `go install` for purposes of installing artifacts in a "canonical" location (that might be problematic).

I'm accustomed to (if not thrilled by) the `autoconf` world where `./configure --prefix=` and friends would set things like this up, the resulting `make`-built executable(s) would have the prefix info internally, and everything could then be magically packaged up.

If there's a similar mechanism for the world of Go, it didn't reveal itself during my searches.


jcburley:
Note that I did try to put off approaching the classpath issue by incorporating `joker.repl` into the existing mechanisms (`core/data` and `std/`). Neither was immediately fruitful: the former might have worked, but I ran into `defmacro` (and friends) not being defined, and thought better of having `repl.joke` embedded as binary code; while the latter hit the problem of `defmacro` not being supported by `generate-std.joke` (and it wasn't immediately obvious to me how to fix that).

So, since `joker.repl` does not seem to need any code changes inside Joker itself, I thought it'd be a good "early adopter" for the classpath-search mechanism, and spent most of my Saturday doing this.

Hope it has at least some salvageable stuff in it!

jcburley:
Note that I "borrowed" more ideas from https://github.com/anmonteiro/lumo than from Clojure itself. This included experimenting with which version of a "class" (library/package) file got loaded when there were multiple choices.

candid82:
I am not thrilled by the idea of classpath in general (sorry, I should have commented on the issue). I may re-consider if I am presented with a really compelling use case, but so far the current mechanism (loading files from current directory tree) has been sufficient. In fact, I myself have not used even that yet: all my Joker scripts are single files.

Also, I would be very hesitant to distribute Joker in a multi file package. I believe the fact that it's a single binary is very appealing to a lot of people and makes them much more likely to try it out. Requiring building or installation step (install.sh) is a non-starter for Joker.

I do find joker.repl useful, but I think it can be implemented using existing embedding mechanism - see core/data/*.joke file and core/gen_data/gen_data.go generator. The mechanism might not be very obvious though. I should probably try to port joker.repl and document the process, unless you'd like to figure it out on your own. That is, if you are still interested in adding joker.repl at all; I understand that my feedback might be disappointing and I really apologize for not commenting on the classpath issue before.

jcburley:
Thanks for your feedback! I realized this PR represented a fairly bold foray into various areas that might be questionable.

I'll try to unpack issues here (mine plus those you raised) and see if it helps us agree on a course of action.

1. I _do_ need to decouple the current working directory (of a running Joker script) from whence libraries ("classes" in Clojure parlance) are loaded. The code I'm rewriting, and the new code I'll be writing soon, includes what would normally be viewed as scripts and/or Unix command-line tools -- and they'll need to share libraries/packages/classes that are system-wide, without requiring any special handling of the current working directory. That doesn't mandate a _specific_ solution, but it does _require_ one.

2. To solve that first point, one either includes all the potential libraries with the executable, or provides a mechanism to do deliver them separately. As you note, there are already two mechanisms to do the former -- but (as I noted above already) I tried both and ran into troubles. However, there's another approach to do the former that I did find when searching for how Go handles artifact installation -- which is to essentially "append" artifacts (raw or even compressed) to the executable, and pull them in on-demand via e.g. an "Asset()" call, after which they'd be treated as if normal (external) files. I'll investigate the feasibility of doing that next.

3. There's generally an expectation (industry-wide) that an executable for a language implementation is decoupled from installation, upgrades, and removal of libraries used by clients of both the language implementation and those libraries. That is not among my immediate use cases, but might become an issue for me down the road; and, if Joker is to become a viable Clojure interpreter, I would think it'd need to support such decoupling, so I don't consider this classpath PR to be wasted effort (should somebody need it) even if we veto it for now.

4. `joker.repl` itself needn't be loaded by default except when running in REPL mode. I think it's useful to have in that mode, but how it gets there (and whether it's loaded by default as with the current libs) does not seem very important. That's a different status than other libraries (especially those not provided by the Joker project itself, like the ones I'm writing for my own needs) will have.

5. This classpath PR offers a new capability that might be attractive to some existing, or (more likely) future, users -- but it also introduces a dependency on the classpath mechanism to provide and install `joker.repl`, which does not seem necessary at this point (per item number 2 above, assuming we can't easily provide it via the existing mechanisms). In particular, this PR can probably drop the new `install.sh` script and the placement of `joker.repl` in the new `share` directory structure (which would be removed from the PR), since I'm confident we can find an alternate way to ship it (and other similar libraries) with the executable.

So I think we can agree that this PR should drop `install.sh` and the `share/` structure and move `joker.repl` to an existing mechanism or, perhaps, the appended-asset approach (if that pans out).

Given that, would you still object to the goal of this PR, to offer (for users who desire it) the ability to drop in, upgrade, and remove libraries without having to change the executable?

(Even if you don't object to that goal, you might have issues with the implementation...but we can take these issues on one at a time for now.)

As an aside: I went down this path when I looked at porting `clojure.test` and saw `clojure.repl/doc` referred to. That got me thinking a `joker.repl` would be a great nice-to-have that should be easy to do, so I worked on that...which (after I couldn't make the `core/data/` nor `std/generate-std.sh` approaches work) led to my thinking hey, this is a good excuse to do the classpath thing, which should be easy to do, so....

Anyway, I think at least 65% of my career productivity (especially including wasted effort!) has been the direct, or indirect, result of my saying to myself "well, I don't actually need to do this right now, but it seems easy enough and I'm thinking about it, so why not?".

jcburley:
Note that one drawback of removing `joker.repl` from this classpath mechanism entirely is that there'll be no "natural" testing of this classpath facility.

Of course, I can (and will) add some tests to the `eval-tests.sh` suite -- but I don't think they can (easily) cover all the functionality corresponding to a `-` as a classpath component, which is tied to the full path of the Joker executable itself.

I don't think any of that is a big deal, but wanted to mention it.

Anyway, `joker.repl` would be removed from this PR to one of its own. As I probably should have done in the first place....

jcburley:
So far I've played around with two different approaches to adding arbitrary files to an executable (which would target Joker, for our purposes):

1. Preprocess them into Go source that gets built with the executable. This is essentially a generalization of how `joker/core/data/*.joke` is handled. I've tried https://github.com/gobuffalo/packr/tree/master/v2 but can't quite get it to work; submitted an issue. It looks like it's fairly well-designed at a glance, but it sure does bring in a lot of dependencies.

2. Append them to the executable after it gets built. This leverages the fact that (on OSes anybody seems to care about) arbitrary data can be appended to executables, and it won't bother the loading; plus, the path to the executable can be discovered in various ways based on the executable. The overhead is thus quite low; the executable is made larger, but that extra data is referenced only if it is actually needed. I've tried https://github.com/YourFin/binappend-cli and it seems to work (plus it supports compressed files), but is rather crude and doesn't appear to be a full-throated implementation. We could just use the `osext` module that it uses and grow our own solution, thus avoiding more than one extra dependency (I think).

Both of these approaches allow Joker to continue to "ship" as a single executable, but with additional libraries/classes/packages that require little or no special handling by any of the Joker code itself, can be fairly large, and add minimal overhead for any Joker user that doesn't reference (load) them.

Besides these approaches and my original approach of shipping files separately, I'm unaware of any other solutions to my main issue, that of needing arbitrary code running in arbitrary directories to `:require` system-wide libraries.

Please give this all some thought. I'll work on other stuff in the meantime -- I don't (yet) have an immediate need for any of this.

jcburley:
Also note that I might be misunderstanding the current mechanisms in re the "stack" of Joker/Clojure packages, or maybe I'm wrong to view them as implementing two of (at least) three potential layers in such a stack in the first place.

I think of `core/data/*.joke` as "ground zero" (the lowest level in the Clojure-code stack) -- code that Joker really can't run without, for whatever reason. `core.joke` clearly belongs there; I don't know (offhand) about the others.

Then, `std/` has the next level up in the stack -- the system/host libraries, so to speak. These interface to Joker internals and/or the underlying Go runtime, and so require special handling (`generate-std.joke`) in many cases.

Above that would be "pure" Joker/Clojure-code libraries. These don't need special help in the Joker executable to perform their functions.

I think `joker.repl` and `joker.test` would belong among those "pure" libraries, but might be wrong.

candid82:
Thanks for digging into this. Looks like there are two separate (but potentially related) questions here: 
1. How to ship additional libraries (not belonging to joker.core) with Joker. Example: joker.repl or joker.test.
2. What's the mechanism for sharing user libraries across the apps.

The first question is easier to answer. As I mentioned above, the existing mechanism (perhaps with some tweaks) for embedding Joker code (core/data/*.joke) should work for things like joker.repl or joker.test, unless I am missing something. I'll give it a try this week to make sure I am not crazy and this actually works.

The second question is a hard one. I did not think it was important enough due to lack of use cases, but I guess it's now time to give it a proper thought.

> The code I'm rewriting, and the new code I'll be writing soon, includes what would normally be viewed as scripts and/or Unix command-line tools -- and they'll need to share libraries/packages/classes that are system-wide, without requiring any special handling of the current working directory.

 As far as I can tell, most ecosystems move away from shared libraries. Node.js [doesn't support](https://nodejs.org/api/modules.html) loading modules from a global location (barring GLOBAL_FOLDERS, which is strongly discouraged, and using symlinks). Python community had to invent [virtualenv](https://virtualenv.pypa.io/en/latest/) to isolate the apps and to avoid sharing dependencies. 
Let me think more about your use case. It would help if you could share a bit more context (perhaps some examples of shared libraries you are envisioning)...

jcburley:
> Thanks for digging into this. Looks like there are two separate (but potentially related) questions here:
> 
> 1. How to ship additional libraries (not belonging to joker.core) with Joker. Example: joker.repl or joker.test.
> 2. What's the mechanism for sharing user libraries across the apps.
> 
> The first question is easier to answer. As I mentioned above, the existing mechanism (perhaps with some tweaks) for embedding Joker code (core/data/*.joke) should work for things like joker.repl or joker.test, unless I am missing something. I'll give it a try this week to make sure I am not crazy and this actually works.

Okay. I might play with it too. In my (perhaps incorrect) view of the world, it seems "obvious" to me that higher-level libraries, that depend on things like `joker.core`, `joker.net`, `joker.os`, and so on, would not be able to live down in the lowest layer that `core/data/` represents (to me).

> The second question is a hard one. I did not think it was important enough due to lack of use cases, but I guess it's now time to give it a proper thought.
> 
> > The code I'm rewriting, and the new code I'll be writing soon, includes what would normally be viewed as scripts and/or Unix command-line tools -- and they'll need to share libraries/packages/classes that are system-wide, without requiring any special handling of the current working directory.
> 
> As far as I can tell, most ecosystems move away from shared libraries. Node.js [doesn't support](https://nodejs.org/api/modules.html) loading modules from a global location (barring GLOBAL_FOLDERS, which is strongly discouraged, and using symlinks). Python community had to invent [virtualenv](https://virtualenv.pypa.io/en/latest/) to isolate the apps and to avoid sharing dependencies.
> Let me think more about your use case. It would help if you could share a bit more context (perhaps some examples of shared libraries you are envisioning)...

I'll get back to you on this -- I want to research what Python et al support (and recommend) over and above adding libraries to their source trees and rebuilding their executables. (I come from a background where executables are fairly rigid, even "sacrosanct", and thus not to be arbitrarily rebuilt or mucked with just to add things like business logic for a given enterprise...but the times they are a-changin'!)

jcburley:
I've glanced at a couple of the links you provided on system-wide shared libraries. They're not _entirely_ what I'm focused on (though in my current use case, they'd do just fine) -- as I've said, I'm also interested in the general use case of language processor LP (this would be the Joker executable in our case) being independent of client code LC needing arbitrary 3rd-party libraries 3PL installed, to be called by LC and processed by LP.

Do you know offhand whether the ClojureScript and/or Clojure communities have addressed these same issues? That could be helpful, because my work in this PR was largely to copy what they do (modulo the fact that we don't have an equivalent to JAR files).

Even though Joker is still somewhat in its infancy, I do think it'd be wise to avoid simply copying an older approach to such issues that maturing technologies are now getting away from.

jcburley:
Also, could you explain the logic in `lib_path__` that removes component names, from the current working directory path, corresponding to the number of components in the current namespace (`*ns*`)? I can see what it does (and experimented with it in my own build), but am not clear as to how that becomes a useful or desirable feature, especially in light of (someday) supporting e.g. `(joker.os/cwd "some-other-directory")`.

In particular, I haven't come across any `os.Chdir()` calls that would traverse downwards as multi-level namespaces are loaded.

I suspect I'm missing something obvious, but would appreciate a "Dummies Guide to Module Loading" or similar at this point!

candid82:
RE: joker.repl. See https://github.com/candid82/joker/commit/8e9545815dc6f443e74facaf1bdca9c49e39ae65 Hopefully this clarifies things a bit. Basically `core/gen_data/gen_data.go` is a mechanism for embedding arbitrary Joker code into the binary. It turns those 'core/data/*.joke' files into `core/a_*.go` files with the content like this:

```go
func init() {
	replData = []byte("\x00...")
}
```

These `*Data` byte slices can then be passed to `processData()` 

I'll comment on classpath / shared libraries later.

jcburley:
Superseded by https://github.com/candid82/joker/pull/152.

candid82:
@jcburley  RE: `lib-path__`. Lib loader assumes that namespaces' names correspond to their location in the filesystem. Specifically, last part of the namespace (after the last dot) should match the file name (without the `.joke` extension) and preceding parts correspond to the path to the file (with dots separating directories).
For example, say you have a directory named `ttt`  with the following structure:
```
├── core.joke
└── utils
    ├── a.joke
    └── b.joke
```
And here is the content of `*.joke` files:

```clojure
;; core.joke
(ns ttt.core
  (:require [ttt.utils.a :refer [a]]))

(a)

;; utils/a.joke
(ns ttt.utils.a
  (:require [ttt.utils.b :refer [b]]))

(defn a []
  (println "I am A")
  (b))

;; utils/b.joke
(ns ttt.utils.b)

(defn b []
  (println "I am B"))
```

Then you can run `core.joke` and get the following (expected) output:

```
I am A
I am B
```

This mechanism is similar to how class loading works on JVM (and how namespace loading works in Clojure). Please note that current working directory doesn't come into play here: `lib-path__` resolves libraries' paths relative to the file currently being executed. 
Does this help?

jcburley:
@candid82 -- the above explanation does nicely explain how `a.b.c` (as a namespace) is translated to `a/b/c.joke` (relative to _something_).

However, I was referring to this logic in `core/procs.go/procLibPath`:

```
	ns := GLOBAL_ENV.CurrentNamespace().Name
	parts := strings.Split(ns.Name(), ".")
	for _ = range parts {
		file, _ = filepath.Split(file)
		file = file[:len(file)-1]
	}
```

This effectively "backs up" the tree, from `$PWD` (when `GLOBAL_ENV.file.Value == nil`), one parent dir per nonterminal component in the current namespace, before then appending e.g. `a/b/c.joke`.

E.g.:

```
$ ./joker
Welcome to joker v0.11.0. Use EOF (Ctrl-D) or SIGINT (Ctrl-C) to exit.
user=> (require 'a.b.c)
<joker.core>:3475:13: Eval error: open /home/craig/go/src/github.com/candid82/joker/a/b/c.joke: no such file or directory
Stacktrace:
[...]
user=> (ns user1.user2)
nil
user1.user2=> (require 'a.b.c)
<joker.core>:3475:13: Eval error: open /home/craig/go/src/github.com/candid82/a/b/c.joke: no such file or directory
Stacktrace:
[...]
user1.user2=>
```

Note how the `joker` component -- the final one in `$PWD` -- is removed during the second `(require 'a.b.c)`, because the current namespace then has two components. If a third was added, `candid82` would then be removed, and so on.

When `GLOBAL_ENV.file.Value != nil`, as I suspect is the main use case for this, that probably makes sense, because then `$PWD` is indeed not involved at all (well, except during the initial file load that makes that value non-`nil`).

I think I'm just puzzled as to whether this behavior is really intended for top-level `(require ...)`, when there's no input filename (e.g. from a REPL or a `--eval` option or similar).

But it's also not clear to me that it's the right thing when loading files, since not all file-loading is necessarily going to be to handle `(require ...)`. E.g. a classic `(load-file ...)` might be involved, in which case any `(require ...)` it does will be subject to the same path-shortening, in that the path of the loaded file would get shortened based on the number of components in the current namespace as of each `(require ...)` it does.

Hope this clarifies my confusion for you!  ;)

jcburley:
BTW, once this is ironed out, if you're okay with it, I'd like to document it in the docstring for `require`. (This thread is hard enough for me to find, nevermind a "casual" user of Joker....)

candid82:
`require` has to use *some* algorithm to find the file that contains requested namespace. Current behavior attempts to mimic Clojure's behavior in the common use case when there is a project directory tree and namespaces require other namespace inside this directory tree. You are right, `GLOBAL_ENV.file.Value != nil` is the main use case. 
I am not really sure what should be the right behavior for `GLOBAL_ENV.file.Value == nil`. Perhaps it should just assume cwd to be the root for all required namespaces, regardless of the current namespace? This would probably make more sense compared to the current behavior. What do you think?
I did not quite understand the problem with `load-file`. Can you provide a specific example?

jcburley:
> `require` has to use _some_ algorithm to find the file that contains requested namespace. Current behavior attempts to mimic Clojure's behavior in the common use case when there is a project directory tree and namespaces require other namespaces inside this directory tree. You are right, `GLOBAL_ENV.file.Value != nil` is the main use case.

Yes, I think that makes sense.

> I am not really sure what should be the right behavior for `GLOBAL_ENV.file.Value == nil`. Perhaps it should just assume cwd to be the root for all required namespaces, regardless of the current namespace? This would probably make more sense compared to the current behavior. What do you think?

It might, especially if cwd can be overridden by whatever Joker offers _a la_ `--classpath` or `*classpath*` or whatever it's named. As I mentioned before, that would allow decoupling of `$PWD` from where the Joker script itself live and where the libraries it (might) load live.

> I did not quite understand the problem with `load-file`. Can you provide a specific example?

I haven't constructed one, and it might be too contrived to worry about (especially if we change the behavior as described above; though maybe that won't be enough to achieve what I'm concerned about here).

But I was thinking something along the lines of a `(load-file "more-code.joke")`, where `more-code.joke` does a `(require ...)` -- or maybe the other way 'round.

I think it's important for the lookup algorithm for all the `require` stuff to be on an entirely separate "track" from `load-file` and friends, in that `load-file` should pertain to `$PWD` (at the time of the file load) and not otherwise be subject to "classpath" searching or anything.

Nor should the location of a file being loaded via `load-file` have any effect on where a library, that it loads via `require`, is searched.

Please let me know whether you want me to play around with this more and come up with some examples and/or propose a design.

jcburley:
> > I think it's important for the lookup algorithm for all the require stuff to be on an entirely separate "track" from load-file and friends, in that load-file should pertain to $PWD (at the time of the file load) and not otherwise be subject to "classpath" searching or anything.
> 
> > Nor should the location of a file being loaded via load-file have any effect on where a library, that it loads via require, is searched.
> 
> Not sure why we would want `load-file` and `require` to behave differently and what problem that would solve. Let's postpone this discussion until someone runs into issues here.
>
> I do agree we should change the behavior for `GLOBAL_ENV.file.Value == nil` case as discussed above.

I've made a note to look into this further and expect to submit a PR down the road -- either to propose a fix or to improve the docstring(s).

PR #140, opened 2018-12-11
Support joker.os/chdir

jcburley:
(Not sure that `MakeError()` is quite right for the long term.)

### DIFF
--- a/core/object.go
+++ b/core/object.go
@@ -1419,6 +1419,13 @@ func (s String) Compare(other Object) int {
 	return strings.Compare(s.S, s2.S)
 }
 
+func MakeError(e error) String {
+	if e == nil {
+		return MakeString("")
+	}
+	return MakeString(e.Error())
+}
+
 func IsSymbol(obj Object) bool {
 	switch obj.(type) {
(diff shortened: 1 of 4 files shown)


candid82:
The route I've chosen for error handling is to throw exceptions. Here is an [example](https://github.com/candid82/joker/blob/master/std/url/url_native.go#L9-L16). Might not be the best choice, but it's better to keep it consistent.

jcburley:
> The route I've chosen for error handling is to throw exceptions. Here is an [example](https://github.com/candid82/joker/blob/master/std/url/url_native.go#L9-L16). Might not be the best choice, but it's better to keep it consistent.

I was about to copy that approach, but thought maybe for `os.Chdir` a "not found" might be, well, less than exceptional.

For now, though, I'll bring it into line with everything else. Might be worth revisiting if native Go types (like `error`) are ever surfaced to Joker code....

jcburley:
Hmm, but what would it return as a value if not an error object? For now I'll have it always return `NIL`.

candid82:
Thanks!

PR #141, opened 2018-12-13
Add joker versions of clojure.template and clojure.walk

jcburley:
These are used by the (forthcoming) `joker.test`. Seemed easier to port than to rewrite around them.

### DIFF
--- a/core/data/template.joke
+++ b/core/data/template.joke
@@ -1,0 +1,55 @@
+;   Copyright (c) Rich Hickey. All rights reserved.
+;   The use and distribution terms for this software are covered by the
+;   Eclipse Public License 1.0 (http://opensource.org/licenses/eclipse-1.0.php)
+;   which can be found in the file epl-v10.html at the root of this distribution.
+;   By using this software in any fashion, you are agreeing to be bound by
+;   the terms of this license.
+;   You must not remove this notice, or any other, from this software.
+
+;;; template.clj - anonymous functions that pre-evaluate sub-expressions
+
+;; By Stuart Sierra
+;; June 23, 2009
+
+;; CHANGE LOG
+;;
+;; June 23, 2009: complete rewrite, eliminated _1,_2,... argument
+;; syntax
+;;
+;; January 20, 2009: added "template?" and checks for valid template
+;; expressions.
+;;
+;; December 15, 2008: first version
+
+
+(ns ^{:doc "Macros that expand to repeated copies of a template expression."
+       :author "Stuart Sierra"}
+  joker.template
+  (:require [joker.walk :as walk]))
+
+(defn apply-template
+  "For use in macros.  argv is an argument list, as in defn.  expr is
+  a quoted expression using the symbols in argv.  values is a sequence
+  of values to be used for the arguments.
+
+  apply-template will recursively replace argument symbols in expr
+  with their corresponding values, returning a modified expr.
+
+  Example: (apply-template '[x] '(+ x x) '[2])
+           ;=> (+ 2 2)"
+  [argv expr values]
+  (assert (vector? argv))
+  (assert (every? symbol? argv))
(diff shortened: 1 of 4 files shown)


jcburley:
Ah, the failing tests here (`flag-tests.sh` complaining repeatedly that `joker.walk` and `joker.template` are unused namespaces) have taught me a bit more about why I had trouble adding `joker.repl` this way.

I'd put off having it auto-referall'ed when running in REPL mode (I noticed Clojure itself did that), so it wasn't being loaded at startup.

Then, starting it off with `(in-ns 'joker.repl)`, as `time.joke` and such did, I got build complaints that e.g. `defn` wasn't defined -- which I'm now getting when I try to fix the above failing check by changing `(ns ...)` to `(in-ns ...)`, even if I add `(use 'joker.core)` (since `use` itself is defined in `joker.core`).

I'm not sure what the right solution is, but perhaps (as suggested by the comment I added in `procs.go`) it's simply to defer loading the data until actually referred to by some code, e.g. via `:require` or similar?

Will investigate that later, if I don't hear back from you first.

jcburley:
This is a "short-term fix", in my view, because it special-cases the two new libraries here. This approach allows `(use '[joker.walk] :reload true)` to work multiple times, which is not the case for `(use '[joker.core] :reload true)`, as there's currently no special-case check for the forced reloading of `joker.core` to come from internal data.

It's not obvious to me which result is correct. In a "development mode" of sorts, someone (like one of us here developers) might want to be able to just drop a new `core.joke` (or `time.joke` or whatever) into the appropriate path and force a reload to pick it, instead of the internal version, up.

If that's the case (i.e. if forced reloading should look only to the filesystem, not any internal version), this "short-term fix" would need to be revised so the special-casing is done only when reloading isn't forced.

Otherwise, we should probably generalize the special-casing here, and add the other `core/data/` and even `std/` namespaces to it, which could have an added benefit of (slightly) quicker startup times.

(I realize `joker.math` is actually a combination of both of the mechanisms, so the generalized approach should support that, or perhaps `joker.math` should be combined into a single `math.joke` somehow.)

candid82:
I reverted your short-term fix and instead fixed the linter for Joker and EDN dialects: dc0d22f7843726c78990912f17e658413bf5f321 and 9ecdaccb8f6cb4aa1dfc94ab32c3b979f59f0f38.

jcburley:
Yes, the short-term fix intentionally treated only the two new namespaces specially. I agree that they should be treated the same as all the other (non-`core`) namespaces -- and it isn't obvious to me whether it's better to load them only on demand, or automatically at startup.

(Might be worth comparing the startup times with and without automatic loading at startup, even though there are still rather few libraries at this point.)

jcburley:
(Note that I added a block comment to this PR before noticing you'd already merged it!)

jcburley:
> In general it might makes sense to only load `joker.core` on startup, and only load other `joker.*` namespaces on demand (after `require` is called). However, this should be implemented for all non-core`joker.*` namespaces, not just for the random subset. And, frankly speaking, I am not sure this is necessary better than loading all `joker.*` namespaces on startup... Clojure itself loads some non-core namespaces on startup. And for scripting purposes it's sometimes convenient to be able to do `joker.os/sh` (or call other `joker.*/*` functions) without having to require those namespaces first.

I wonder if it's possible (without the code getting too hairy) to do "on-the-fly" loading upon seeing something like `(joker.foo/somefunc ...)` when `joker.foo` is not loaded but is available to be loaded?

That'd give us the ease-of-use that we covet, without costing much of anything in terms of startup time, especially as more libraries are added.

jcburley:
Note that `(use '[joker.repl])` does not work outside of REPL mode. Surprisingly enough, that complicates testing the `joker.test` module I'm working on...!

PR #144, opened 2018-12-15
defmulti and Friends

jcburley:
Tried my hand at adding these, so `joker.test` wouldn't look quite so horribly different than `clojure.test` (but I did get it working that way).

Went with "pure" Joker/Clojure code, instead of Clojure's approach of using an internal type.

After doing this, looked at ClojureScript, and it's somewhat similar to my approach here, except it's built on `deftype`, which Joker doesn't have (but we should add soon IMO).

Not sure whether this is acceptable as a stopgap (it's incomplete, but should be easy to fill out if you agree on the approach), but it should probably be changed later on to use an underlying type like ClojureScript or Clojure, rather than an anonymous meta map.

In the meantime, it'll unblock a reasonably clean `joker.test` being added.

Note that I made a few guesses at cleaning up the lint mechanism, but might have gotten some things wrong there.

Please let me know what you think of this approach. I'm happy to revise it, as I get time, once the direction is clear, if it is deemed not mergeworthy at this time. (I'd really like to finish up the `gostd2joker` work, which I think will require a number of improvements to Joker, and want `joker.test` and a bunch of decent tests to already exist before I undertake that work.)


jcburley:
Sorry, deleted this branch (instead of another one not ready for PR) by accident; restored and reopened.

candid82:
I think this approach is fine. I personally think multimethods are rarely useful, but it'd be good to support them for better compatibility with Clojure. As you pointed out, this will make porting Clojure code (like clojure.test) easier. I actually like the fact that introducing multimethods did not require deeper changes in Joker's runtime.

I'll try to provide more detailed feedback over the weekend.

core/data/core.joke
```diff
@@ -4189,3 +4196,190 @@
...
   []
   (joker-version__))
+
+(defn- check-valid-options
+  "Throws an exception if the given option map contains keys not listed
+  as valid, else returns nil."
+  [options & valid-keys]
+  (when (seq (apply disj (apply hash-set (keys options)) valid-keys))
+    (throw
+      (ex-info ; was: (IllegalArgumentException.
```
candid82: `was:...` comments are probably unnecessary.

core/data/core.joke
```diff
@@ -4189,3 +4196,190 @@
...
+        (apply str "Only these options are valid: "
+          (first valid-keys)
+          (map #(str ", " %) (rest valid-keys))) {}))))
+
+;;multimethods
+(defn- space-str
```
candid82: We should try to minimize the number of non-Clojure functions in the core namespace, even if they are private, to reduce the probability of name collisions. If necessary, such functions should be named with `__` suffix, so `space_str__` in this case. Although this function is only used in one place, so should probably be inlined. 
Also, the implementation could be simpler: `(apply str (interleave (repeat " ") more))`

core/data/core.joke
```diff
@@ -4189,3 +4196,190 @@
...
+  "Quick and dirty version of joker.string/join with a single space as the separator, because joker.core probably shouldn't depend on joker.string"
+  ([] "")
+  ([a] (str " " a))
+  ([a & more] (str " " a (apply space-str more))))
+
+(defn- multifn-nomatch
```
candid82: See the comment above. This function can be inlined.

core/data/core.joke
```diff
@@ -4189,3 +4196,190 @@
...
+  [name args]
+  (throw (ex-info (str "No method in multimethod for dispatch value: (" name (apply space-str args) ")") {})))
+
+(defn- new-multifn
```
candid82: I would prefer the name `multimethod__` for this function.

core/data/core.joke
```diff
@@ -4189,3 +4196,190 @@
...
+          (apply method args)))
+      {:name name :dispatch-fn dispatch-fn :default default :ns *ns* :method-table mfatom})))
+
+(defn- multifn-swap-method-table-vals!
+  [multifn ^Callable f & args]
+  (let [mfm (meta multifn)
+        mfatom (:method-table mfm)]
+    (apply swap-vals! mfatom f args)))
+
+(defn- multifn-reset
```
candid82: This and below private functions can and should be inlined in their public counterparts IMO.

core/data/core.joke
```diff
@@ -4189,3 +4196,190 @@
...
+      (fn [& args]
+        (let [method (get @mfatom
+                          (apply dispatch-fn args)
+                          (get @mfatom
+                               default
+                               (fn [& args] (multifn-nomatch name args))))]
+          (apply method args)))
+      {:name name :dispatch-fn dispatch-fn :default default :ns *ns* :method-table mfatom})))
+
+(defn- multifn-swap-method-table-vals!
```
candid82: This function should be named with `__` suffix.

core/data/core.joke
```diff
@@ -4189,3 +4196,190 @@
...
+
+(defmacro defmethod
+  "Creates and installs a new method of multimethod associated with dispatch-value. "
+  {:added "1.0"}
+  [multifn dispatch-val & fn-tail]
+  `(multifn-swap-method-table-vals! ~multifn assoc ~dispatch-val (fn ~@fn-tail)))
+
+(defn remove-all-methods
+  "Removes all of the methods of multimethod."
+  {:added "1.2"
```
candid82: `:added` should be `1.0` everywhere, as it refers to the version of Joker, not Clojure. Hopefully one day there will be version 1.0...

jcburley: Thanks for the thorough review! Mainly regarding inlining, note that I used Clojure's own `defmulti` code as a template for this effort, and kept the separation of public-function definitions from private-function implementations I saw there (and later saw that ClojureScript copied as well).

I don't mind combining them as suggested, but first want to make sure you're okay with moving the code base further afield from upstream.

(If `defmulti` was a rapidly moving target, i.e. a hotbed of innovation, I'd be more concerned about doing that!)

core/data/core.joke
```diff
@@ -1333,6 +1333,13 @@
          (let [~form temp#]
            ~@body)))))
 
+(defn class
+  "Returns the Class of x; currently this is the underlying type, as Go classes are not yet supported."
```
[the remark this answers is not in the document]
jcburley: Using the same wording as for `type`, so: "Returns the Type of x."

core/data/core.joke
```diff
@@ -1333,6 +1333,13 @@
...
+  {:added "1.0"
+   :static true}
```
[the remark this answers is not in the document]
jcburley: Removed everywhere.

core/data/core.joke
jcburley: Removed everywhere.

core/data/core.joke
jcburley: Nice! Hadn't used `interleave` before....

core/data/core.joke
jcburley: Done.

core/data/core.joke
jcburley: Hmm, before or after the `!`?

core/data/core.joke
jcburley: I'm okay with that, but note that I used Clojure's own defmulti code as a template for this effort, and kept the separation of public-function definitions from private-function implementations I saw there (and later saw that ClojureScript copied as well).

I don't mind combining them as suggested, but first want to make sure you're okay with moving the code base further afield from upstream.

(If defmulti was a rapidly moving target, i.e. a hotbed of innovation, I'd be more concerned about doing that!)

core/data/core.joke
jcburley: See below (https://github.com/candid82/joker/pull/144#discussion_r243853933).

core/data/core.joke
candid82: The difference is that both Clojure and ClojureScript call methods of `IMultiFn` (or `MultiFn`), rather than private functions. This makes sense since the functionality is encapsulated in special types. Since Joker's implementation doesn't use special types, I think this addition layer is unnecessary.

core/data/core.joke
jcburley: Done.

core/data/core.joke
jcburley: Yes, definitely not needed, and the code is easier to read that way.

Done!

jcburley:
Once this is merged, I'll circle back to getting `joker.test` in as a PR, along with some tests.

I'm not sure any of the unimplemented multimethod functions (like `remove-method`) are needed -- `joker.test` wouldn't appear to need them -- but somebody can Issue that if necessary. Shouldn't be terribly difficult to add....

candid82:
Great! Thanks!

PR #146, opened 2018-12-17
Report min:max arity range when handy

jcburley:
E.g.:

```
user=> ({:a 1 :b 2})
<repl>:9:1: Eval error: Wrong number of args (0) passed to `fn'; expects 1:2
user=>
```

Not sure `min:max` notation is the right thing here; maybe `min through max`, or `min-max` (though that looks like a subtraction as much as a range to many) would be better?


core/object.go
```diff
@@ -456,10 +457,22 @@ func PanicArity(n int) {
 	panic(RT.NewError(fmt.Sprintf("Wrong number of args (%d) passed to %s", n, name)))
 }
 
+func rangeString(min, max int) string {
+	if min == max {
+		return strconv.Itoa(min)
+	}
+	return strconv.Itoa(min) + ":" + strconv.Itoa(max)
```
candid82: Maybe `from X to Y`? Or, better yet, if Y - X <= 2, something like `X, X+1 or Y`? Also, it'd be good to special case if Y = 999 (or maybe change that magic number to `MaxInt`) to handle cases like [this](https://github.com/candid82/joker/blob/master/std/os/a_os.go#L122); it could say `X or more` in this case. What do you think?

jcburley:
Glad you like it! Incorporated suggested improvements. The one that's kinda wordy ("between MIN and MAX, inclusive") currently won't ever occur, AFAICT. (I was a tech writer decades ago, and was taught the importance of such precision -- in particular, "3 to 5" could be interpreted as "3 or 4".)

candid82:
Looks great! I think sum256dir belongs to another PR though :)

jcburley:
> Looks great! I think sum256dir belongs to another PR though :)

Sorry -- juggling too many branches on my laptop! Fixed.

PR #148, opened 2018-12-18
 Produce more-gofmt-compatible output

jcburley:
Scratching another itch here, as I move `gostd2joker` into the Joker source tree and continue work on it, and want to ensure my (Go) code is properly formatted without hassle....


candid82:
Thanks!

PR #150, opened 2018-12-19
Make CircleCI run tests properly in forks

conormcd:
In forks of this project, `CIRCLE_PROJECT_USERNAME` is not `candid82` and so references in code or elsewhere to `candid82/joker` will fail. The solution is to work in and check out directly to the directory in `GOPATH`. This simplifies the config and makes it possible to test this project from a fork.


conormcd:
Example run: https://circleci.com/gh/conormcd/joker/3

PR #151, opened 2018-12-19
clojure.core/proxy exposes 'this to its body

conormcd:
Tweak the known-macros map to take note of the fact that `'this` is bound within the body of a proxy.


candid82:
Thanks!

PR #152, opened 2018-12-20
Add Classpath Support

jcburley:
Per discussion in https://github.com/candid82/joker/pull/139, resurrecting this (after merging), as Joker doesn't yet have a proper solution for 3rd-party/system-wide libraries ("proper" excluding solutions like "drop libraries into Joker source tree and rebuild Joker" and "always run Joker from a specific directory that has libraries in expected places").

Still not sure this is the _best_ thing, but it's the closest (I can think of) to industry standards and practices, if not to the latest industry aspirations....


candid82:
@jcburley I still think this is not ideal, but I realize the "proper" way to solve the problem (in my mind, at least) would probably involve creating a full-fledged package manager, which is a pretty big project that may take very long time to materialize. So I am OK with adding "classpath" as a stop-gap. The implementation needs to be reduced to bare bones though.
- Not sure I understand the need for both `classpath` and `classdir`. `classdir` seems redundant to me.
- There is no need for `*classpath*` var.
- Let's rename "--classpath" to "--lib". Joker doesn't have classes, neither does Go.
- Let's leave `--lib` option undocumented for now.

How does this sound to you?

jcburley:
After stepping away, I realized I hadn't made it clear what `$GOROOT/lib` represented.

It's just the idea of a place that would hold vendor/3rd-party artifacts for Go-built binaries.

I actually don't think `$GOROOT/lib` is the right place for adding our own stuff, because it gets into a subtree that is currently entirely the domain of the Go language/runtime itself.

But something like `$GOROOT/vendor/github.com/candid82/joker/lib/` seems less intrusive,  more self-explanatory, and perhaps worthy of official adoption (by the Go ecosystem) down the road.

jcburley:
Another thought: modifiable vars like `*classpath*` are really necessary only because the (built-in) functions using them aren't "hookable".

If they were (somehow), such that the "default" hooks would be to use whatever was canonical, but user-provided hooks would be passed that canonical thing and return what should be used instead, there'd be no need to support a modifiable form of that thing.

That doesn't solve the issue of (Joker) code reading/reporting the canonical thing without having to attempt a (hooked) operation, but it does show another approach to solving the testability issue.

candid82:
My suggestions above are reflection of the fact that I consider this feature (in whatever form it ends up being implemented) experimental and subject to change / removal. Hence the desire to add as little as possible to the code and surface area and leave it undocumented.

1. Naming. `classpath` is probably fine. An alternative would be `libpath` I think, but it doesn't matter that much.
2. `classdir` and canonical place for libraries. As I understand this is not really required if `classpath` is implemented as one can always provide that "canonical" place as `--classpath` argument. This might be a nice addition down the road but is not essential now and therefore should probably be stripped out of the initial implementation. If we do add it in the future, it should not be coupled with Go's environment variables (`$GOROOT`) as Joker doesn't (and should not) require Go to run. Something like `$JOKERROOT` might make sense, but again, not now.
3. `*classpath*` var. I did not see the use case for it (beyond "might be useful", which doesn't cut it in this context), but you mentioned testing. Could you please elaborate on how this would help with testing?

jcburley:
> My suggestions above are reflection of the fact that I consider this feature (in whatever form it ends up being implemented) experimental and subject to change / removal. Hence the desire to add as little as possible to the code and surface area and leave it undocumented.

Hmm, okay, I can see that argument!

> 1. Naming. `classpath` is probably fine. An alternative would be `libpath` I think, but it doesn't matter that much.

Agreed. I lean a bit towards `classpath` (even though it's misleading) only because it probably fits better with likely user expectations -- but that's only because I'm defining "likely user" as "Clojure/ClojureScript user looking for a good linter and/or quick-starting, low-overhead interactive experience", rather than "someone coming from a non-Clojure background who wants to play around with a lightweight Clojure-like environment that might possibly become a viable shell" and so might find the `classpath` terminology confusing.

> 2. `classdir` and canonical place for libraries. As I understand this is not really required if `classpath` is implemented as one can always provide that "canonical" place as `--classpath` argument. This might be a nice addition down the road but is not essential now and therefore should probably be stripped out of the initial implementation.

Yes, I think if we're not going to try to leverage Go's environment variables and directory structure, we can punt `classdir` entirely (at least for now).

If we do add it in the future, it should not be coupled with Go's environment variables (`$GOROOT`) as Joker doesn't (and should not) require Go to run. Something like `$JOKERROOT` might make sense, but again, not now.

Okay.

> 3. `*classpath*` var. I did not see the use case for it (beyond "might be useful", which doesn't cut it in this context), but you mentioned testing. Could you please elaborate on how this would help with testing?

I think it'll be easier to have a single invocation of Joker, along with a "static" tree of `.joke` files acting as libraries, and simply change `*classpath*` around to confirm different effects for `require` and such (to make sure directories are searched in the correct order, reloading/forced-reloading works, etc.) than to have distinct tests (in distinct directories), each of which must re-invoke Joker with a `--classpath` (or equivalent) option to set up the desired environment.

But I haven't written much in the way of tests for this yet, so currently it's a bit of future-proofing on my part.

Given your concerns about documenting this mechanism, I guess `*classpath*` could (like `*loaded-libs*`) be made private?

core/environment.go
```diff
@@ -54,6 +56,18 @@ func (env *Env) SetEnvArgs(newArgs []string) {
 	}
 }
 
+func (env *Env) SetClassPath(cp string) {
+	cparray := filepath.SplitList(cp)
```
candid82: Here and below, names should be camel case: `cpArray` etc.

core/environment.go
```diff
@@ -54,6 +56,18 @@ func (env *Env) SetEnvArgs(newArgs []string) {
...
+	cpvec := EmptyVector
+	for _, cpelem := range cparray {
+		cpvec = cpvec.Conjoin(String{S: cpelem})
```
candid82: It's not used widely in the codebase, but there is `MakeString` function.

core/environment.go
```diff
@@ -79,6 +93,9 @@ func NewEnv(currentNs Symbol, stdin io.Reader, stdout io.Writer, stderr io.Write
 			:minor and/or :major, bugfix releases will increment :incremental.`, "1.0"))
 	res.args = res.CoreNamespace.Intern(MakeSymbol("*command-line-args*"))
 	res.SetEnvArgs(os.Args[1:])
+	res.classPath = res.CoreNamespace.Intern(MakeSymbol("*classpath*"))
+	res.classPath.Value = NIL
+	res.classPath.isPrivate = true
```
candid82: Sadly, this is not enough: the var's meta needs to contain private keyword as well. See https://github.com/candid82/joker/blob/master/core/procs.go#L1550-L1558.

main.go
```diff
@@ -368,6 +369,11 @@ func parseArgs(args []string) {
 	stop := false
 	missing := false
 	noFileFlag := false
+	if v, ok := os.LookupEnv("JOKER_CLASSPATH"); ok {
```
candid82: I'd really like to remove `JOKER_CLASSPATH` from this initial implementation. I feel like it's here mostly because it's "cute" rather than because it's serving any particular purpose.

core/procs.go
```diff
@@ -1388,14 +1389,44 @@ var procLoadFile Proc = func(args []Object) Object {
...
+		}
+		f, err = os.Open(filename)
+		if err == nil {
+			break
+		}
+		if s == "" {
+			canonicalErr = err
+		}
 	}
+	if canonicalErr == nil {
```
candid82: Error handling here seems more complicated than it should be. What's the difference between `err` and `canonicalErr`? Don't we want to panic on `err` either way (since `canonicalErr` is either equal to `err` or `nil`)?

core/data/core.joke
```diff
@@ -3264,6 +3264,11 @@
   [lib]
   lib)
 
+(defn lib-path
```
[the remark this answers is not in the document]
jcburley: Privatized.

core/data/core.joke
```diff
@@ -3453,13 +3461,21 @@
   *loaded-libs*)
 
 (defn load-file
-  "Loads code from file f"
+  "Loads code from file f. Does not protect against recursion."
   {:added "1.0"}
   [^String f]
   (load-file__ f))
 
+(defn load-lib-from-path
```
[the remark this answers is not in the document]
jcburley: Privatized.

core/environment.go
jcburley: Fixed.

main.go
jcburley: It indeed has a purpose (avoids need to create a wrapper that serves as a shell, passing `--classpath` in to the Joker executable, for e.g. when ssh'ing in to run a script) but not yet one I actually need, though I anticipate needing it fairly soon.

There are some interesting online discussions about `$CLASSPATH` that raise issues about it, though they pertain to Clojure/Java, which is not intended to serve as a shell (or offer the quick startup that we're after with Joker) -- and, of course, Joker doesn't need a "build tool" to run its programs. E.g.:

https://stackoverflow.com/questions/3639747/environment-variable-classpath-clojures-classpath-why
https://www.ibm.com/support/knowledgecenter/en/ssw_ibm_i_73/rzaha/classpth.htm

By contrast, Planck (ClojureScript) has its own `$PLANCK_CLASSPATH`, though there's not much discussion as to why or the pros and cons of using it:

http://planck-repl.org/dependencies.html
http://blog.fikesfarm.com/posts/2017-01-22-planck-road-ahead.html

Let me know whether you'd like me to remove this to a (potential) future PR where we can discuss this further. It's not much extra work, and I don't mind scoping down this PR for now.

jcburley:
Besides the unresolved conversations, I plan to add some tests to this PR soon.

So it's not ready for merging IMO.

jcburley:
Some tests have been added. Once this PR is merged (in some form), I'll add an Issue on the fact that the new test breaks if `input.joke` starts off with (e.g.) `(ns user.my-ns)`, which works fine in `lumo`.

core/procs.go
candid82: OK, I think this could be simplified to 
```go
PanicOnErr(canonicalErr)
PanicOnErr(err)
```
However, I wonder if we should provide more user friendly error message here. Something like `Failed to load namespace X. Tried locations: f1, f2...`

main.go
candid82: Yep, let's remove this for now.

core/data/core.joke
```diff
@@ -3264,6 +3264,11 @@
   [lib]
   lib)
 
+(defn- lib-path
```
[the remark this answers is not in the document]
jcburley: Done.

candid82:
Thank you very much!

PR #155, opened 2018-12-22
Run gofmt

jcburley:


jcburley:
Might consider blocking PRs that don't pass `go fmt -n`...? Or maybe that's overkill at this point?

candid82:
Blocking PRs feels like an overkill at this point. I do recommend integrating your editor with gofmt. I use GoSublime for Sublime Text and it runs gofmt on file save. I am sure plugins exist for other editors as well.

Thanks!

PR #164, opened 2019-01-09
Fix Test Comparisons on Windows

jcburley:
I'm not 100% sure this is the right approach. It'll prevent editing of (most) Joker source files via NotePad on Windows (since NotePad doesn't handle Unix-style line terminators correctly). I'm not sure whether there'll be other issues.

But it avoids requiring tools like `run-eval-tests.joke` to "manually" strip out carriage returns in expected outputs (they don't seem to show up in generated output, at least not in Git Bash) before comparing them.

Since I don't currently have a non-Git-Bash, barebones Joker development environment on a Windows machine, it's hard to predict what effect this might have on such environments; but they're currently probably rather broken anyway due to lack of a `run.bat` and other equivalents for shell scripts.

While doing this work (which was rather painful, due to Git Bash not behaving as documented with respect to changing `.gitattributes` on the GitHub site and elsewhere), `docs/joker.core.html` kept getting changed. Arguably `generate-docs.joke` has a bug wherein it outputs embedded `\r` (CR) characters in multiline docstrings (which are read in by Joker when files are in DOS format), resulting in sequences like `foo\r<br>...`; this approach avoids, but does not fix, that bug.

If moving to Unix line endings for most/all files is unacceptable, I'll try the other approach of having the various test scripts replacing `\r\n` with `\n` in expected (`slurp`'ed) outputs.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,17 +1,5 @@
 # Set the default behavior.
-* text=auto
+* text eol=lf
 
-# Unix line endings on .txt files.
-*.txt text eol=lf
-
-# Unix line endings on .go files.
-*.go text eol=lf
-
-# Unix line endings on .html files.
-*.html text eol=lf
-
-# Unix line endings on .joke files.
-*.joke text eol=lf
-
-# Unix line endings on .tmpl files.
-*.tmpl text eol=lf
+# DOS line endings on .bat files.
+*.bat text=auto
--- a/.gitattributes
+++ b/.gitattributes
@@ -3,3 +3,6 @@
 
 # Unix line endings on .txt files.
 *.txt text eol=lf
+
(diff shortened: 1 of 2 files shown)


candid82:
I don't develop on Windows so unix line endings everywhere works for me. Cannot we simply do `* text eol=lf` at the top of the file though?

jcburley:
> I don't develop on Windows so unix line endings everywhere works for me. Cannot we simply do `* text eol=lf` at the top of the file though?

Sounds good to me, let's try it.

Still carving out an exception for `*.bat` files though, so we avoid whoever adds `run.bat` and similar tearing their hair out wondering why it isn't working.

candid82:
Thanks!

PR #166, opened 2019-01-15
Support Go1.12 Beta 2

jcburley:
This should work for both Go 1.12 and Go 1.11.

Also note that `set -e` has been replaced with `trap` per best practices, mentioned here:

https://stackoverflow.com/a/4384381/8869495

(The script wasn't always stopping upon `go vet` failure, probably because `set -e` or `trap` is needed as well in the `build` function.)

### DIFF
--- a/run.sh
+++ b/run.sh
@@ -5,9 +5,9 @@ build() {
 
     go generate ./...
 
-    go vet -all ./
+    go vet -all ./...
 
-    [ -n "$SHADOW" ] && go vet -all "$SHADOW" ./ && echo "Shadowed-variables check complete."
+    [ -n "$SHADOW" ] && go vet -all "$SHADOW" ./... && echo "Shadowed-variables check complete."
 
     go build
 }
--- a/run.sh
+++ b/run.sh
@@ -7,7 +7,13 @@ build() {
 
     go vet -all ./...
 
-    [ -n "$SHADOW" ] && go vet -all "$SHADOW" ./... && echo "Shadowed-variables check complete."
+    if [ -n "$SHADOW" ]; then
+        go vet -all "$SHADOW" ./... && echo "Shadowed-variables check complete."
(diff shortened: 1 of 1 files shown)


candid82:
Thank you for looking into this! I don't really like that `run.sh`, which is supposed to be super simple script that just executes a few commands, is growing more and more complex. I admit I am biased here because I dislike (to put it softly) Bash syntax and try to avoid it  for anything except very simple linear scripts. I think I'd prefer to keep `run.sh` as is until the time comes to actually upgrade to Go 1.12 (when it's out of Beta). At that point it'll be OK to drop "official" support for Go 1.11 for dev environment, which will hopefully keep `run.sh` simple.

jcburley:
Sounds good to me!

On January 15, 2019 11:55:05 PM EST, Roman Bataev <notifications@github.com> wrote:
>Thank you for looking into this! I don't really like that `run.sh`,
>which is supposed to be super simple script that just executes a few
>commands, is growing more and more complex. I admit I am biased here
>because I dislike (to put it softly) Bash syntax and try to avoid it 
>for anything except very simple linear scripts. I think I'd prefer to
>keep `run.sh` as is until the time comes to actually upgrade to Go 1.12
>(when it's out of Beta). At that point it'll be OK to drop "official"
>support for Go 1.11 for dev environment, which will hopefully keep
>`run.sh` simple.
>
>-- 
>https://github.com/candid82/joker/pull/166#issuecomment-454651679

-- 

candid82:
@jcburley I addressed this by removing shadow check altogether: https://github.com/candid82/joker/commit/a7fb34e027487bd9b2cfe813597787a2cca438d0

jcburley:
Note that shadow check _did_ find one of the little bugs my other PR fixes! So we might want to restore that after fully moving to Go 1.12....

Issue #157, opened 2019-01-02
(load-file) Does Not Throw for Parse Errors -- Silent Failure


jcburley:
E.g.:

```
$ cat bad-load
foo
"hey"
(println "hey there")
$ ./joker
Welcome to joker v0.11.0. Use EOF (Ctrl-D) or SIGINT (Ctrl-C) to exit.
user=> (load-file "bad-load")
bad-load:1:1: Parse error: Unable to resolve symbol: foo
nil
user=> (try (load-file "bad-load") (catch Error e (println "failed")))
bad-load:1:1: Parse error: Unable to resolve symbol: foo
nil
user=>
```

Compare:

```
Clojure 1.9.0
user=> (load-file "bad-load")
CompilerException java.lang.RuntimeException: Unable to resolve symbol: foo in this context, compiling:(/home/craig/go/src/github.com/candid82/joker/bad-load:1:1)
user=> (try (load-file "bad-load") (catch Exception e (println "failed")))
failed
nil
user=>
```

This is a problem for the (forthcoming) `joker.test` and related `tests/run-eval-tests.joke` (per https://github.com/candid82/joker/issues/131), because it can lead to a test silently failing, in that there'd be error output but no (straightforward) detection by the test runner such that a nonzero exit code would be returned by the script.

candid82:
Fixed by https://github.com/candid82/joker/commit/ed7f4035ab48f93e463bc0757ef1b63a6e046363

PR #167, opened 2019-01-20
Ensure load-file throws on parse error

jcburley:
Confirms https://github.com/candid82/joker/issues/157 is fixed.


candid82:
Thanks!

Issue #170, opened 2019-01-21
Duplicate Case Constants Undetected


jcburley:
Unlike Clojure, Joker won't detect this:

```
(case "hey" :this :that :this :whatever :oy)
```

E.g. Clojure reports `Duplicate case test constant: :this`.

As an interpreter, it doesn't seem terribly important; but as a linter, it might be a good idea to do.

candid82:
Fixed by https://github.com/candid82/joker/commit/229ce29a059300018c82626beab353d6c9611f78

Issue #169, opened 2019-01-21
(= 1 1N) Evaluates to False


jcburley:
...whereas `(= 1N 1)` evaluates true, as expected.

(Discovered while porting `control.clj`.)

candid82:
Fixed by https://github.com/candid82/joker/commit/28f6933aeff5ac9a59502dcac4fc05c48300b80c

PR #171, opened 2019-01-21
Port clojure.test-clojure.control

jcburley:
This exposed a bug in `joker.test`, also fixed herein.

I'm not sure `test_helper.joke` is in the right place. The way `require` currently works (with respect to the multilevel namespaces, as in `*ns*`, used by this test infrastructure) complicated my attempts to do something a little more elegant. For now, it seems to work.

### DIFF
--- a/tests/eval/control.joke
+++ b/tests/eval/control.joke
@@ -14,7 +14,7 @@
 
 (ns joker.test-clojure.control
   (:require [joker.test :refer [deftest is are testing]]
-            [tests.test-helper :refer :all]))
+            [tests.test-helper :refer [exception]]))
 
 ;; *** Helper functions ***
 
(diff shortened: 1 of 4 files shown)


jcburley:
This also exposed #169 and #170, plus a potential issue (search `control.joke` for "boom") that might not be worth worrying about.

tests/eval/control.joke
```diff
@@ -0,0 +1,443 @@
...
+
+; Author: Frantisek Sodomka, Mike Hinchey, Stuart Halloway
+
+;;
+;;  Test "flow control" constructs.
+;;
+
+(ns joker.test-clojure.control
+  (:require [joker.test :refer [deftest is are testing]]
+            [tests.test_helper :refer :all]))
```
candid82: Since Joker, unlike Clojure, doesn't convert dashes to underscores when loading files, let's rename `test_helper` (the file and the namespace) to `test-helper`. Also, please don't use `:refer :all` as it makes the code unlintable.

jcburley: Done.

candid82:
Thanks!

PR #172, opened 2019-01-22
Add a tiny section with some advice from PR feedback

jcburley:
There're probably more suggestions that could be added, but this is a modest start.


candid82:
Thanks! I am going to move this section down below though since its audience is very small.

PR #173, opened 2019-01-23
Delete trailing whitespace

jcburley:
Somehow missed seeing this earlier...yet they show up as red boxes on my Mac laptop, so they should have been obvious (guess I didn't `git diff` the full PR itself before submitting).


candid82:
Thanks!

Issue #165, opened 2019-01-09
Linter Test types-1 Hangs on Windows


jcburley:
This is in Git Bash. I modified `run-linter-tests.joke` to print each test before running it, and it hangs after printing `Running types-1`.

Haven't yet tracked down which particular form (if any), within that input file, is causing the problem.

jcburley:
Seems to be an actual hang (not a CPU hog).

Also, can't reproduce this directly via

```
./joker --lint tests/linter/types-1/input.clj
```

but can repro via:

```
./joker  -e '(joker.os/sh "./joker" "--lint" "tests/linter/types-1/input.clj")'
```

Commenting out the number of lines until the uncommented (active) lines count reaches anywhere in the 24-36 range seems to avoid the issue -- it evidently has nothing to do with any particular line of input.

Also, it doesn't seem to matter whether `types-1` runs earlier -- I made a copy of it, named `a-types-0`, and it shows the same issues (right away, since it sorts first in the list of tests to run).

So this appears to be some strange bug involving Joker shelling out to Joker. Given the arbitrary behavior, it could just as easily be a Go (Windows-runtime) bug, or a Git Bash or Windows bug, as a Joker bug -- in fact, I don't suspect Joker at this point.

jcburley:
FYI, this still happens with Go 1.11.4 (which I installed on my Windows dev box 2019-01-16; previously it ran 1.11.2).

PR #176, opened 2019-01-24
Read from forked process stdout and stderr asynchronously

jcburley:
Fixes https://github.com/candid82/joker/issues/165.


candid82:
That's a really great catch! Thank you!

PR #178, opened 2019-01-25
Port clojure.set

jcburley:
Wanted `joker.data` too, but that needs `defprotocol` or a more-significant port effort. But this seems useful and was easy.


jcburley:
(Testing this is done manually, because of the instability of output. It seems to work fine on Ubuntu and Mac OS X.)

candid82:
Thanks!

PR #180, opened 2019-02-21
Add a few more java.lang classes to the linter

conormcd:
The following don't need to be imported and hence shouldn't fail linting.

- java.lang.BigInteger
- java.lang.Thread$State
- java.lang.Thread$UncaughtExceptionHandler


candid82:
Thanks!

PR #184, opened 2019-03-04
Allow passing char to some joker.string functions

mrichards42:
Clojure isn't entirely consistent with this, but it is common to allow passing a char as a secondary argument in functions that operate on strings (e.g. `join`, `replace`, `index-of`).

Adding a `Stringable` pseudo-type seemed like the most straightforward way to accomplish this, but it does feel a little awkward, since all the other types have generated `Ensure` and `Assert` functions. I'd be open to doing this another way if you'd prefer.

### DIFF
--- a/core/procs.go
+++ b/core/procs.go
@@ -94,6 +94,10 @@ func ExtractString(args []Object, index int) string {
 	return EnsureString(args, index).S
 }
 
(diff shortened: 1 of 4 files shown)


candid82:
I like it. Thanks!

PR #185, opened 2019-03-07
Add joker.string triml and trimr

mrichards42:
Same as trim-left and trim-right, but these are the clojure names


candid82:
Thanks!

PR #186, opened 2019-03-07
Add joker.tools.cli

mrichards42:
This is the current [master](https://github.com/clojure/tools.cli/blob/9e7850caffff45816bec821456345b8268fa72dd/src/main/clojure/clojure/tools/cli.cljc) of clojure.tools.cli and its [test](https://github.com/clojure/tools.cli/blob/9e7850caffff45816bec821456345b8268fa72dd/src/test/clojure/clojure/tools/cli_test.cljc) (I didn't bother including the legacy test).

A couple of minor changes were made to exception code. clojure.tools.cli also uses `string/trimr` which is added in #185, but I'd be fine renaming the call in code to `trim-right`.


core/data/tools_cli.joke
```diff
@@ -0,0 +1,706 @@
...
+                           (filter (complement :flag))
+                           (mapcat :switches)
+                           (into #{}))
+        ;; Preserve double-dash since this is a pre-processing step
+        largs (take-while (partial not= "--") args)
+        rargs (drop (count largs) args)
+        [opts largs] (tokenize-args required-opts largs)]
+    (concat (mapcat rest opts) largs rargs)))
+
+(defn cli
```
candid82: I think we should not bring over legacy API.

core/data/tools_cli.joke
```diff
@@ -0,0 +1,706 @@
...
+             (if (contains? (set ids) id)
+               [m ids errors]
+               [(assoc m id (f (first %))) ids errors]))
+           % default-fns))
+        (#(let [[m ids errors] %]
+            (if no-defaults
+              [(select-keys m ids) errors]
+              [m errors]))))))
+
+(defn ^{:added "0.3.0"} make-summary-part
```
candid82: All `:added` tags need to be changed to `^{:added "1.0"}`. I know they are meaningless until v1.0 actually comes out, but that's the convention for pre-v1.0 API.

candid82:
Thanks! I pointed out a couple of minor things before this can be merged. Also, `joker.string/trimr` is now in master, so tests should pass once you merge master into your branch.

core/data/tools_cli.joke
mrichards42: Sounds good to me 👍

mrichards42:
Thanks for the feedback! I rebased onto master to make sure the ci tests would run

candid82:
Awesome, thanks!

PR #187, opened 2019-03-07
exit 1 when eval errors

mrichards42:
The tests in #186 failed, but circleci reports that they passed, because joker exits 0 in all cases when evaluating code. I'm not sure 1 is the correct exit code (since there are a lot of exit codes in main.go), but I don't think it should be 0.

Specifically, the [eval tests](https://circleci.com/gh/candid82/joker/599) failed since I forgot to push the `gen_data.go` file in that PR.


candid82:
Thanks!

PR #194, opened 2019-03-21
Populate *command-line-args* When Documenting It

jcburley:
An empty VectorSeq is more consistent than nil, especially since the
docs are generated by a Joker script, normally run with no arguments,
that uses the type of *command-line-args* to describe its type in the
docs.

### DIFF
--- a/core/environment.go
+++ b/core/environment.go
@@ -49,7 +49,11 @@ func (env *Env) SetEnvArgs(newArgs []string) {
 	for _, arg := range newArgs {
 		args = args.Conjoin(MakeString(arg))
 	}
-	env.args.Value = args.Seq()
+	if args.Count() > 0 {
+		env.args.Value = args.Seq()
+	} else {
+		env.args.Value = NIL
+	}
 }
 
 func (env *Env) SetClassPath(cp string) {
--- a/core/environment.go
+++ b/core/environment.go
@@ -49,11 +49,7 @@ func (env *Env) SetEnvArgs(newArgs []string) {
 	for _, arg := range newArgs {
 		args = args.Conjoin(MakeString(arg))
 	}
-	if args.Count() > 0 {
-		env.args.Value = args.Seq()
-	} else {
-		env.args.Value = NIL
-	}
+	env.args.Value = args.Seq()
 }
 
 func (env *Env) SetClassPath(cp string) {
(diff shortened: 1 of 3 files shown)


candid82:
Clojure docs say

> A sequence of the supplied command line arguments, or nil if
none were supplied.

So it seems to me the current behavior is "per spec".

jcburley:
> Clojure docs say
> 
> > A sequence of the supplied command line arguments, or nil if
> > none were supplied.
> 
> So it seems to me the current behavior is "per spec".

Hmm, well that's a bit strange, but it's best to be compatible with Clojure IMO.

Seems like something should be done about [this](https://candid82.github.io/joker/joker.core.html#*command-line-args*), though:

```
  <h3 id="*command-line-args*">*command-line-args*</h3>
  <span class="var-type Nil">Nil</span>
```

I.e. the type is _not_ truly  _always_`Nil`.

Passing a "dummy" argument to Joker, when generating the docs, would yield a `var-type` of `VectorSeq`, which I think would be an improvement, albeit a modest one given that it would still not be entirely correct and the docstring explains what is returned anyway.

I'll update the PR to do that, in case you find that acceptable. Another approach I can think of is to provide metadata with the variable such that `generate-docs.joke` uses that metadata as an override for what its `type-name` function currently does.

jcburley:
A little background: on my `gostd` fork, I decided to scratch an itch by creating a `cleanup.sh` script to undo the effects of running `tools/gostd/gostd`, which creates many files, and modified a few, in the Joker source tree. Trying to do this somewhat elegantly (e.g. without just clobbering modified files via `git reset --hard HEAD`, since I sometimes modify them for other reasons in my work), I decided to teach `generate-docs.joke` to filter out the `go.std.*` namespaces when a `--no-go` command-line argument was passed in, which the `cleanup.sh` script does.

That's when I noticed the change in `docs/joker.core.html` to the type for `*command-line-args*`.

There are a couple of other `Nil`-typed variables, both in `joker.test`, that I might look into later, after we decide what (if anything) to do about this (fairly trivial) issue.

candid82:
I think the metadata approach is more flexible so I went ahead and fixed it that way: https://github.com/candid82/joker/commit/d44ca9fd66eb34853377c373e5374f2bb07fa929 Thank you for bringing this to my attention.

jcburley:
Nice! Though your approach has left some very long lines in the docs -- would it work to replace `\n` with `<br>\n` instead of just `<br>`?

candid82:
Yes, I think this makes sense. Thanks for the suggestion! https://github.com/candid82/joker/commit/2916d7708714fe6bbdb2aabaf7d457c9d000171b

Issue #196, opened 2019-03-22
Common REPL functions missing


arichiardi:
Hello folks,

Great great project, I am excited to try it out and use it for some scripts.

My non-urgent issue is that while patching [`inf-clojure`](https://github.com/clojure-emacs/inf-clojure/pull/160) so that it can connect to a `joker` process I noticed that some of the common REPL functions are missing. Here is the list:
  
* `joker.repl/dir` 
* `joker.repl/source`
* `joker.repl/apropos`

It would be great to have them in if possible.

jcburley:
Agreed! When I first created `joker.repl`, it was a quick-and-dirty thing to get just what I needed in. Wasn't sure anyone else would find it helpful. Intended to circle back and do this sometime...am looking at it right now, to see how easy it is. Thanks!

candid82:
`dir` and `apropos` are now implemented thanks to @jcburley. `source` is not really feasible with current implementation and it certainly does not seem important enough to warrant major re-design.

PR #197, opened 2019-03-22
Add dir and apropos functions

jcburley:
These were easy to do. The source functions need further thought.

Partially addresses: https://github.com/candid82/joker/issues/196


candid82:
Thanks a lot!

arichiardi:
Awesome thank you! I will patch `inf-clojure` to support these when a new release is out :smile:

PR #202, opened 2019-03-28
Fix typo in README.md

traviswsims:
Minor correction – changed "The code it totally valid" to "The code is totally valid."

Thanks for the hard work!


candid82:
Thanks!

Issue #203, opened 2019-04-03
-1 for --hashmap-threshold not accepted


lread:
If I try the example in the README:
```
joker --hashmap-threshold -1 -e "(pprint (read))"
```
I get the following result:
```
Error: Missing argument for '--hashmap-threshold' option
```
I am running joker v0.12.1 installed via brew on macOS Mojave v10.14.4

jcburley:
This looks like my bug, as I rewrote the command-line processing a few months ago.

Looks like my code treats `-1` (negative one) as an option and so doesn't consider it as a valid argument to a preceding option.

Will push a PR soon.

PR #204, opened 2019-04-04
Allow negative numbers as arguments

jcburley:
Fixes: https://github.com/candid82/joker/issues/203

### DIFF
--- a/main.go
+++ b/main.go
@@ -360,8 +360,13 @@ var (
 	noReadline         bool
 )
 
+func isNumber(s string) bool {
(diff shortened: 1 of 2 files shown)


jcburley:
For now, you can probably achieve the same effect via `--hashmap-threshold 2147483647`.

candid82:
Thanks! I hit this issue before and used the workaround you mentioned. It's nice to have it fixed though.

lread:
Thanks so much for fixing this!

PR #205, opened 2019-04-10
[main] Add ability to set classpath with an environment variable.

charlesg3:
For some of the work I'm doing, I have scripts that simply specify the joker command to run. They rely on external libraries for which I would like to specify a class path externally using an enviornment variable.

This PR allows you to specify the class path using "JOKER_CLASSPATH=dir1:dir2" syntax just like you would with the "-c" option on the command line.

It would definitely be a help to my use of joker if this was in master.

Much thanks for all of your work!

### DIFF
--- a/main.go
+++ b/main.go
@@ -557,6 +557,16 @@ var runningProfile interface {
 	Stop()
 }
 
+func getEnv(key string) string {
+	for _, e := range os.Environ() {
+		pair := strings.Split(e, "=")
+		if pair[0] == key {
+			return pair[1]
+		}
+	}
+	return ""
+}
+
 func main() {
 	InitInternalLibs()
 	ProcessCoreData()
@@ -578,6 +588,9 @@ func main() {
 	}
 
 	parseArgs(os.Args)
+	if classPath == "" {
+		classPath = getEnv("JOKER_CLASSPATH")
+	}
 	GLOBAL_ENV.SetEnvArgs(remainingArgs)
 	GLOBAL_ENV.SetClassPath(classPath)
 


jcburley:
Hmm, I think I still have a similar change on a branch of my fork -- after it was decided to not provide this functionality along with the (undocumented) --classpath option I added a few months ago.

Maybe your use case (if not your PR) will be more convincing than my tentative one!

On April 10, 2019 1:06:45 PM EDT, Charles <notifications@github.com> wrote:
>For some of the work I'm doing, I have scripts that simply specify the
>joker command to run. They rely on external libraries for which I would
>like to specify a class path externally using an enviornment variable.
>
>This PR allows you to specify the class path using
>"JOKER_CLASSPATH=dir1:dir2" syntax just like you would with the "-c"
>option on the command line.
>
>It would definitely be a help to my use of joker if this was in master.
>
>Much thanks for all of your work!
>You can view, comment on, or merge this pull request online at:
>
>  https://github.com/candid82/joker/pull/205
>
>-- Commit Summary --
>
>  * [main] Add ability to set classpath with an environment variable.
>
>-- File Changes --
>
>    M main.go (13)
>
>-- Patch Links --
>
>https://github.com/candid82/joker/pull/205.patch
>https://github.com/candid82/joker/pull/205.diff
>
>-- 
>You are receiving this because you are subscribed to this thread.
>https://github.com/candid82/joker/pull/205

-- 

charlesg3:
Oh wow -- I hope that the classpath support is retained...

I use it so that I can have scripts related to an individual project along with a "library" that is relevant across many projects. This is the first time in a while that I've been able to conveniently leverage shared code for my ops infrastructure (eliminating a lot of duplicated code).

jcburley:
See my old PR, which I just hastily resurrected:

https://github.com/candid82/joker/pull/206

I think the only salient difference (besides the API used to access the environment) is that mine also changes the docs, but:

- I'm not 100% sure those (changed) docs are correct even with the PR, as I think we were in the midst of making changes in how the classpath stuff actually behaved around the same time I removed `JOKER_CLASSPATH` env var support to a different branch.
- Since `--classpath` (`-c`) isn't (yet) documented, @candid82 might not want to document the new env var yet.

jcburley:
@candid82 -- If you were to add this functionality, I think documenting it would be of vital importance, because the possibility would then exist that someone trying to debug a Joker system would not know why or from whence code was being loaded, because they wouldn't know to look for a `JOKER_CLASSPATH` env var.

(With `--classpath` aka `-c`, at least they'd presumably notice the undocumented option on the command line....)

candid82:
Hm... I still dislike the idea of classpath. On the other hand, I see that people start to use Joker for scripting and *some* way of sharing code has to exist. I'll merge James's PR to add support for `JOKER_CLASSPATH`, but I am still hesitant to document this functionality and make it "official".

My vision for Joker deps management is more like this: https://clojurians.slack.com/archives/C9VURUUNL/p1554511813028900?thread_ts=1554412026.019400&cid=C9VURUUNL

@charlesg3  If Joker supported something like `(require '<namespace> :from <url>)`, where `<url>` is, say, a github url, would that work for your use case (and eliminate the need for `--classpath`)?

charlesg3:
I agree about the `:from` approach. I started to actually go down that direction but then picked up a few issues, curious to hear your thoughts on them:

- My joker libs are in a private repo, so it's a little challenging to get the `git` commandline tool to get them from github.com because you still need to specify some credentials to get to the private repos... that felt like more of a rabbit-hole, but I didn't explore that fully.

- For production systems, you need `releases` instead of `snapshots`, as such it's nice to be able to specify a tag or sha (as in `deps.edn` with clojure). Unfortunately this will start to look a little messy on every include, especially if it needs to be specified in every file

The advantage is that hopefully you could also specify `http(s)://` urls and perhaps even file urls... but again when this spreads across several files in several projects, bumping the version becomes tedious and the require's start to look messy

In either case, I think there's a compelling argument to have the `:from` support on requires. I'm interested in adding the support if we can come to conclusion about how to do it nicely.

charlesg3:
Perhaps we should move this discussion to an issue and close this PR as the support is provided by #206

jcburley:
> Perhaps we should move this discussion to an issue and close this PR as the support is provided by #206

Could even be a Wiki page here. (We've hashed out a few things on Slack in the past, but I think it's better to do stuff like this openly, instead of on a proprietary system.)

candid82:
@jcburley, @charlesg3 Thank you for your feedback! I created https://github.com/candid82/joker/issues/208 to continue the discussion. Interested in your thoughts on the proposal there.

Issue #200, opened 2019-03-24
Assumes 64-bit host


SevereOverfl0w:
Not much of a go expert, but I believe this is caused by https://github.com/candid82/joker/blob/244abd81d51881f5e19d7306429d6a9f5762188a/core/array_map.go#L24 and https://github.com/candid82/joker/blob/244abd81d51881f5e19d7306429d6a9f5762188a/main.go#L438

Where the host is not a 64-bit one, `int` will be smaller than 64-bits.

Specific error is: `./main.go:438:24: constant 9223372036854775807 overflows int`

SevereOverfl0w:
By defining HASHMAP_THRESHOLD as int64 and type casting around comparisons, I seem to be able to get it working on 64-bit hosts. I'm not familiar with go, so I won't open a PR for this, but feel free to use it if it's appropriate.

```
diff --git a/core/array_map.go b/core/array_map.go
index 3d9b95d..6e03cfa 100644
--- a/core/array_map.go
+++ b/core/array_map.go
@@ -21,7 +21,7 @@ type (
 )
 
 var (
-	HASHMAP_THRESHOLD int = 16
+	HASHMAP_THRESHOLD int64 = 16
 )
 
 func EmptyArrayMap() *ArrayMap {
@@ -169,7 +169,7 @@ func (m *ArrayMap) Assoc(key Object, value Object) Associative {
 		res.arr[i+1] = value
 		return res
 	}
-	if len(m.arr) >= HASHMAP_THRESHOLD {
+	if int64(len(m.arr)) >= HASHMAP_THRESHOLD {
 		return NewHashMap(m.arr...).Assoc(key, value)
 	}
 	res := m.Clone()
diff --git a/core/eval.go b/core/eval.go
index 42b7ff2..49ade93 100644
--- a/core/eval.go
+++ b/core/eval.go
@@ -204,7 +204,7 @@ func (expr *VectorExpr) Eval(env *LocalEnv) Object {
 }
 
 func (expr *MapExpr) Eval(env *LocalEnv) Object {
-	if len(expr.keys) > HASHMAP_THRESHOLD/2 {
+	if int64(len(expr.keys)) > HASHMAP_THRESHOLD/2 {
 		res := EmptyHashMap
 		for i := range expr.keys {
 			key := Eval(expr.keys[i], env)
diff --git a/core/read.go b/core/read.go
index 199fd72..1a03ca8 100644
--- a/core/read.go
+++ b/core/read.go
[...]

PR #207, opened 2019-04-11
Fix builds for 32-bit hosts

jcburley:
Fixes https://github.com/candid82/joker/issues/200 with only slight modifications to the proposed changes (to ensure the command-line value is parsed as an `int64` in the first place).

Tested via:

```
$ GOOS=linux GOARCH=386 go build
$ file joker
joker: ELF 32-bit LSB executable, Intel 80386, version 1 (SYSV), statically linked, not stripped
$ ./joker
Welcome to joker v0.12.2. Use EOF (Ctrl-D) or SIGINT (Ctrl-C) to exit.
user=> ^D
$ ./all-tests.sh
[...]
Ran 59 tests containing 511 assertions.
0 failures, 0 errors.
All three tests passed.
$
```

### DIFF
--- a/core/array_map.go
+++ b/core/array_map.go
@@ -21,7 +21,7 @@ type (
 )
 
 var (
-	HASHMAP_THRESHOLD int = 16
+	HASHMAP_THRESHOLD int64 = 16
 )
 
 func EmptyArrayMap() *ArrayMap {
@@ -169,7 +169,7 @@ func (m *ArrayMap) Assoc(key Object, value Object) Associative {
 		res.arr[i+1] = value
 		return res
 	}
-	if len(m.arr) >= HASHMAP_THRESHOLD {
+	if int64(len(m.arr)) >= HASHMAP_THRESHOLD {
 		return NewHashMap(m.arr...).Assoc(key, value)
 	}
 	res := m.Clone()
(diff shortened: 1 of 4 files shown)


candid82:
Thanks! I think I've made a few assumptions about `int` being 64-bit in other places, so I won't be surprised if Joker crashes or doesn't work correctly on 32-bit systems though.

cellularmitosis:
Thanks so much for this fix!  This has made it possible for me to run a bit of Clojure on some old ARM hardware.

On an [NSLU2](https://en.wikipedia.org/wiki/NSLU2) (233MHz ARMv5, 32MB RAM), `joker` startup time is under 5 seconds:

```
$ echo '(+ 1 1)' | time joker 
Welcome to joker v0.12.2. Use EOF (Ctrl-D) or SIGINT (Ctrl-C) to exit.
2
3.98user 0.32system 0:04.31elapsed 99%CPU (0avgtext+0avgdata 16332maxresident)k
0inputs+0outputs (0major+4494minor)pagefaults 0swaps
```

For comparison, Clojure startup time is about 60x slower (a large component of which is waiting on swapfile I/O, I'm sure):

```
$ time clj -e '(System/exit 0)'
75.71user 46.69system 4:40.35elapsed 43%CPU (0avgtext+0avgdata 18384maxresident)k
4000256inputs+0outputs (41298major+231803minor)pagefaults 0swaps
```

```
$ clj -e '(clojure-version)'
"1.10.0"
```

```
$ java -version
java version "1.8.0_201"
Java(TM) SE Embedded Runtime Environment (build 1.8.0_201-b09, headless)
Java HotSpot(TM) Embedded Client VM (build 25.201-b09, mixed mode)
```

candid82:
Nice!

PR #210, opened 2019-04-18
[deps] Add initial support for dependency management.

### LINKED ISSUE #208: Dependency managament
See https://github.com/candid82/joker/pull/205 for some context.

Joker currently supports undocumented `--classpath` command line parameter (and corresponding `JOKER_CLASSPATH` environment variable). It works the same way as Java's classpath and provides a way to use shared "libraries". There is a number of issues with classpath approach though:
- It doesn't provide delivery mechanism. Joker will try to load libraries from the specified location(s), but how those libraries get there is up to the user.
- It doesn't provide explicit version management. If many programs use the same shared library, it becomes tricky to update it without breaking anything.
- It's external to programs' source code. One cannot tell which external dependencies will be loaded at runtime just by looking at the code.

Ideally, Joker's dependency management system should be
- simple, both conceptually and implementation wise
- explicit: it should be easy to tell which dependencies come from where just by looking at the source code
- built-in: there should not be a separate tool or command to run to pull dependencies

I tentatively proposed `:from <url>` option for `(require)`, but @charlesg3 brought up a valid concern about this approach:

	> For production systems, you need releases instead of snapshots, as such it's nice to be able to specify a tag or sha (as in deps.edn with clojure). Unfortunately this will start to look a little messy on every include, especially if it needs to be specified in every file

To address the above concern, I think Joker should allow to declare all dependencies in one place, similarly to `deps.end` or lein's `project.clj`, except it should be part of the source code.

Here is how it could work.

There is `*ns-sources*` built-in var. It's a map where keys are namespace names (possibly with wildcards) and values are "sources", which are instructions on how and where to load those namespaces from. For example:

```clojure
{"com.test1.common.*" {:url "/var/lib/joker/test1/common"} ; local file system
 "com.test2.common.*" {:url "test2/common#v2.1.0"} ; github
 "com.test3.common.*" {:url "git+ssh://git@github.com:test3/common.git#v1.0.0"} ; git
}
```

The types of different urls have been borrowed from node.js's npm.
When Joker needs to load a namespace (i.e. while executing `(require ...)`), it first consults *ns-sources* map. If the namespace's name matches one of the keys, Joker will try to load the namespace from the corresponding source. Except for the simplest case of local file system urls, the loaded namespaces will have to be cached somewhere. The list of supported urls could be extended in the future, or perhaps the loading mechanism could be made pluggable. For example, user could provide their own loading function in addition to the url. In addition to `*ns-sources*` var, there should be `(ns-sources <map>)` function to conveniently set the var. That would normally be the first statement in a Joker script. For example:

```clojure
(ns-sources
  {"com.test.common.*" {:url "test/common#v2.0.0"}})

(ns com.test.do-something
  (:require [com.test.common.foo :as foo]))

(println (foo/bar))
```

charlesg3:
Fixes #208


tests/deps/deps.joke
```diff
@@ -0,0 +1,15 @@
...
+  (-> (os/args)
+       (second)
+       (str/split #"/")
+       (butlast)
+       (concat ["deps" "lib"])
+       ((fn [x] (apply str (interpose "/" x))))))
+
+(ns-sources
+  {"test-local.*" {:url lib-dir}
+   "test-http.*" {:url "https://s3.us-east-2.amazonaws.com/tech.public.data/"}})
```
candid82: Tests should not depend on external urls. For now, I'd just remove this check from the test. No tests is better the unstable tests.
I wonder how much of `procLibPath` could be rewritten in Joker. Briefly looking through the code, it should be doable. Then we could mock `joker.http/send` in the tests. This is definitely out of scope for this PR.

core/deps.go
```diff
@@ -0,0 +1,50 @@
...
+	"net/http"
+	"os"
+	"path/filepath"
+	"regexp"
+	"strings"
+)
+
+func externalHttpSourceToPath(lib string, url string) (path string) {
+	home, _ := os.LookupEnv("HOME")
+	localBase := filepath.Join(home, ".joker-deps", "deps", strings.SplitN(url, "//", 2)[1])
```
candid82: Joker already uses `.jokerd` directory for linting related files. I think it makes sense to use `.jokerd/deps` for dependencies.

core/data/core.joke
```diff
@@ -3202,6 +3202,11 @@
     :doc "True while a verbose load is pending"}
   *loading-verbosely* false)
 
+(defonce ^:dynamic
+  ^{:private true
+    :doc "A map from namespace reg-exps to the source to load the external dependency from"}
+  *ns-sources* {})
```
candid82: `*ns-sources*` and `ns-sources` need to be un-mapped [here](https://github.com/candid82/joker/blob/master/core/data/linter_cljx.joke#L382) for linting purposes.

deps-tests.sh
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env bash
+
+./joker tests/run-deps-tests.joke "$@"
```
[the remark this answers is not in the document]
charlesg3: Yep -- integrated into the eval test suite.

tests/deps/deps.joke
charlesg3: - http test removed for now
- good point about serving the file from local.

charlesg3: I thought about implementing more in Joker and started down that path. The trouble I ran into that made me go this way instead was based on dependencies of different things being declared where the `load-lib__` code was being done. It seemed cleaner in the end to just drop down into go and do the resolution there.

If you think otherwise, I can push further on going down the path of doing more in Joker.

core/data/core.joke
charlesg3: When I unmapped `*ns-sources*` at that location I ended up getting the following in other tests:
```
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
```

I can provide full backtrace if that helps, but it wasn't obvious what the error was here. I didn't see any other ear-muffed variables being unmapped at that location.

charlesg3:
@candid82 -- I believe your comments have now been addressed. Thank you for the feedback, please let me know if there's anything else I can do to make this more suitable for merge.

As an aside, I've already started to leverage some of this functionality in some projects that I'm working and it's significantly helping in terms of following DRY principles by allowing me to library-ify a lot of my scripting for dev-ops.

:+1:

tests/deps/deps.joke
candid82: Yeah, bootstrapping may be tricky here. I'll look into it later, for now Go implementation will do.

core/data/core.joke
candid82: I'll look into it later, no big deal.

candid82:
Thanks a lot, @charlesg3, merging!

PR #218, opened 2019-05-02
Create shadow.sh script

jcburley:
This provides the shadowed-variable checking that `run.sh` used to do, but no longer does, keeping that script simple.

This new script supports Go 1.12 as well as previous releases; 1.12 support requires installation of the `shadow` tool to work (see the diagnostic produced for more info).

If this script is found sufficiently worthy, I suppose it could be invoked by `run.sh` in the `build()` function. In the meantime, it's available for developers changing Joker code.

### DIFF
--- a/shadow.sh
+++ b/shadow.sh
@@ -1,0 +1,19 @@
+#!/usr/bin/env bash
+
+# Check for shadowed variables.
+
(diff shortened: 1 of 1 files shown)


candid82:
This is great, thank you James!

PR #227, opened 2019-06-04
Add joker.time/parse and add-date

mrichards42:
A couple extra functions in the joker.time namespace. I didn't get this far, but it would open up the possibility of adding an `#inst` data reader in the future.


candid82:
Thanks!

PR #233, opened 2019-06-12
Pass actual Stdin object to os.exec commands

rwstauner:
I was trying to do something like

```clojure
(defn get-pass [prompt]
  (try
    (stty "-echo")
    (get-input prompt)
    (finally
      (stty "echo")
      (println))))
```

and couldn't because the `:stdin :pipe` wasn't really a terminal.

I added tests for string and :pipe and then made the change to pass the object directly and added some tests for that, though it will only work when there really is a tty.

If you agree with the concept I have some design questions for you:
- do you see any reason to keep the previous :pipe behavior? I believe this works the same, but it could be a different keyword if there was any distinction.
- i added 3 different tests for the tty behavior, all of which are a little weird.  do you prefer one over the others (or have another suggestion)?

I appreciate your feedback.  Joker is really fantastic!  ❤️  Thanks!

### DIFF
--- a/tests/eval/os.joke
+++ b/tests/eval/os.joke
@@ -1,0 +1,8 @@
+(ns joker.test-joker.os
+  (:require [joker.os :as os]
+            [joker.test :refer [deftest is]]))
+
+(deftest exec-pipe
+  (if (= (get (os/env) "TTY_TESTS") "1")
+    (is (= 0 (:exit (os/exec "stty" {:args ["echo"] :stdin :pipe}))))
+    (println "Skipping tty tests (STDIN is not a tty)")))
--- a/tests/run-eval-tests.joke
+++ b/tests/run-eval-tests.joke
@@ -37,7 +37,8 @@
       (println (str "Running test in subdirectory " test-dir)))
     (let [dir (str "tests/eval/" test-dir "/")
           filename "input.joke"
-          res (joker.os/sh-from dir exe filename) ; Someday add: :in "input.txt" (no :in support yet)
+          stdin (slurp-or (str dir "stdin.txt") :pipe)
+          res (joker.os/exec exe {:dir dir :args [filename] :stdin stdin})
           out (:out res)
           err (:err res)
           rc (:exit res)
--- a/eval-tests.sh
+++ b/eval-tests.sh
@@ -1,3 +1,6 @@
 #!/usr/bin/env bash
 
+# Let test suite know to try.
+[[ -t 0 ]] && export TTY_TESTS=1
+
 ./joker tests/run-eval-tests.joke "$@"
--- a/std/os/os_native.go
+++ b/std/os/os_native.go
@@ -71,8 +71,14 @@ func sh(dir string, stdin io.Reader, name string, args []string) Object {
 	PanicOnErr(err)
 	stderrReader, err := cmd.StderrPipe()
 	PanicOnErr(err)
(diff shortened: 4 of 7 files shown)


rwstauner:
It looks like circleci does provide a tty so the tests do run there.

tests/eval/os.joke
```diff
@@ -0,0 +1,9 @@
+(ns joker.test-joker.os
+  (:require [joker.os :as os]
+            [joker.test :refer [deftest is testing]]))
+
+(deftest test-exec
```
candid82: It's not consistent in the code base, but I prefer not to use `test-` prefix. `deftest` is clear enough indication that it's a test :). Also, I try to keep tests flat with no nested `testing`. When possible, each `testing` section should be a separate test, makes it easier to run individual tests. So this test could be named `exec-pipe` and `testing` could be removed from it.

tests/eval/stdin-string/input.joke
```diff
@@ -0,0 +1,3 @@
+(ns stdin-string-test)
+
+(println "|" (read-line))
```
candid82: This test is probably an overkill.

tests/eval/stdin-tty/input.joke
```diff
@@ -0,0 +1,12 @@
+(ns stdin-tty-test
```
candid82: This seems to duplicate `test-exec` above (using a different method of testing). I think it makes sense to use `deftest` method when possible, so this test can be removed.

tests/run-eval-tests.joke
```diff
@@ -37,7 +37,8 @@
       (println (str "Running test in subdirectory " test-dir)))
     (let [dir (str "tests/eval/" test-dir "/")
           filename "input.joke"
-          res (joker.os/sh-from dir exe filename) ; Someday add: :in "input.txt" (no :in support yet)
+          stdin (slurp-or (str dir "stdin.txt") :pipe)
+          res (joker.os/exec exe {:dir dir :args [filename] :stdin stdin}) ; Someday add: :in "input.txt" (no :in support yet)
```
candid82: Looks like the comment is now obsolete and should be removed?

rwstauner: Yes, since the functionality is there, but i wasn't sure how literal (or old) the comment was.

rwstauner:
Thanks for the feedback!  I appreciate your time.  New version pushed!

tests/run-eval-tests.joke
jcburley: > Looks like the comment is now obsolete and should be removed?

That was my comment. Idea there was to detect whether `input.txt` existed in the test's source directory and, if so, automatically set it to be on `stdin` via an `:in` (or now `:stdin`) keyword.

This would simplify at least one existing test, I think, so the test itself wouldn't need to open the file on startup.

IIRC, my intent was to have the file -- not a "slurp'ed" copy -- open on `stdin`. Looks like this new code might do the latter?

Also, I like the `stdin.txt` name better than `input.txt`, _especially_ if the file (not a slurp'ed copy) is open on `stdin`.

candid82: Looks like slurp works fine for this use case and doesn't require additional support from `joker.os/exec`, so let's go with approach this for now.

candid82:
Thanks a lot, @rwstauner!

Issue #237, opened 2019-06-26
eval/stdin-pipe Fails On Linux amd64 With Empty Output


jcburley:
So far I've reduced this to a simple one-liner that shows the string not appearing in the output when passed via `stdin`, while it works when the string is provided directly via the `:stdin` keyword:

```
$ ./joker -e '(println (joker.os/exec "cat" {:stdin :pipe}))' <<< "this is a test"
{:success true, :exit 0, :out , :err }
$ ./joker -e '(println (joker.os/exec "cat" {:stdin "this is a test"}))'
{:success true, :exit 0, :out this is a test, :err }
$
```

This is on my Ubuntu server:

```
Linux dove 4.4.0-142-generic #168-Ubuntu SMP Wed Jan 16 21:00:45 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux
go version go1.11.7 linux/amd64
```

It doesn't appear to be a shell issue or anything else fundamental like that:

```
$ ./joker -e "(read-line)" <<< "this is a test"
"this is a test"
$
```

(it works fine on my Mac OS X.)

jcburley:
Seems to be something to do with buffer flushing? E.g.:

```
$ ./joker -e '(println (count (:out (joker.os/exec "cat" {:stdin :pipe}))))' <<< "this is a test"
0
$ ./joker -e '(println (count (:out (joker.os/exec "cat" {:stdin :pipe}))))' < ./joker
14193658
$ ls -l ./joker
-rwxrwxr-x 1 craig craig 14193658 Jun 26 10:39 ./joker
$
```

jcburley:
I'm not yet convinced this is a bug in Joker itself.

jcburley:
Upgrading to Go 1.11.11 (and rebuilding) doesn't seem to help.

jcburley:
Built and tested a short Go program using nearly-identical code as that in Joker, but couldn't reproduce the problem.

```
$ ./small-stdin-pipe-bug cat <<< "this is a test"
exit=0 outlen=15 errlen=0
$
```

[main.go.txt](https://github.com/candid82/joker/files/3332679/main.go.txt)

jcburley:
I think the next step(s) might include running strace on that small Go program and on the Joker example and comparing them.

(I don't think either the Go debugger or gdb will be of much help, based on past experience, but might give them both a try to figure this out.)

jcburley:
[...]

PR #238, opened 2019-06-27
Fix stdin pipe so output reliably captured

jcburley:
Fix https://github.com/candid82/joker/issues/237 by simply using the built-in `cmd.Stdout` and `cmd.Stderr` fields to hold buffers that will be written to (asynchronously) such that `cmd.Wait()` knows to wait for such writing to complete before tearing things down.

### DIFF
--- a/std/os/os_native.go
+++ b/std/os/os_native.go
@@ -8,7 +8,6 @@ import (
 	"os/exec"
 	"strings"
 	"syscall"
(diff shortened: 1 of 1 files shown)


candid82:
@jcburley sorry for the radio silence, I was on vacation. Thank you very much for fixing this tricky bug and simplifying the code along the way!

Issue #243, opened 2019-08-04
Can't build on armv7 (Raspberry Pi, Go 1.11)


rcarmo:
```bash
pi@quad:~/Library/Go/src/github.com/candid82/joker $ ./run.sh --version 
# command-line-arguments
gen/gen_types.go:88:20: undefined: strings.ReplaceAll
core/object.go:2: running "go": exit status 2
```
I'm guessing this is because `ReplaceAll` seems to be in 1.12, just leaving this here because Raspbian `buster` is still shipping the older version. and someone else might search for this.

rcarmo:
Upgrading to Go 1.12 I now get this when running `go install`:

```bash
# github.com/candid82/joker/std/time
std/time/a_time.go:13:24: constant 3600000000000 overflows int
std/time/a_time.go:17:26: constant 60000000000 overflows int
```

candid82:
@rcarmo I updated Go version in build requirements. 
As for the second issue, 32-bit platforms are not officially supported. In hindsight I should have used `int64` instead of `int` to avoid issues like this, but oh well. If you are really interested in building Joker on a 32-bit platform, try commenting out the code that exposes those int vars in `time.joke`.
I'll see how much work it would be to migrate to `int64` everywhere, but that's definitely not a quick fix.

jcburley:
Is this a different problem (versus just a new instance of the same problem) fixed in https://github.com/candid82/joker/pull/207?

If not, the fix might be limited in scope (haven't checked this myself just yet), and further problems might be avoided by adding a 32-bit-host build to the CI test.

jcburley:
> Is this a different problem (versus just a new instance of the same problem) fixed in #207?
> 
> If not, the fix might be limited in scope (haven't checked this myself just yet), and further problems might be avoided by adding a 32-bit-host build to the CI test.

Ah, having now spent a few minutes looking into this, I see it's a different problem. That earlier PR of mine fixed only internal code; the problem here is the new interface exposed to Joker code, and Joker currently supports only `Int` (Go's `int`, which is 32-bit on 32-bit systems) and `BigInt` (from Go's `big` library), but not a type corresponding to Go's `int64`.

Submitted https://github.com/candid82/joker/pull/244 as one approach to fix this. Been awhile since I've had my head deep into Joker code, so apologies in advance for evident rust or other nonsense.

candid82:
Fixed by https://github.com/candid82/joker/pull/244 Thanks @jcburley!

Issue #174, opened 2019-01-23
Should "Bool" Be "Boolean" Instead, and "Int" "Integer"?


jcburley:
Clojure calls it `Boolean`. Not sure why it's different in Joker....

candid82:
Could have been Boolean... I think I just named it after Go `bool` type, similar to `Int`/`int`, although I did not follow this convention for other types like `Double` or `Char`. I don't think it matters honestly.

candid82:
I do prefer shorter names if it's not at the expense of expressiveness, and `bool` is widely used / understood abbreviation.

jcburley:
It's widely understood outside of Clojure-land, but Clojure using `Boolean` means Clojure code also uses `Boolean`. E.g. `^Boolean` or `:tag Boolean`, such as in `set.clj` in the Clojure source tree.

Seems kinda strange to have a different name for a common type, when the language being implemented is Clojure, just because it's common to use other names in other languages (like the implementation language here).

This is also an issue (to the extent it's much of an issue at all) for `Int` versus `Integer`, hence the Subject modification.

I'm not sure why Clojure chose the wordy spelling. I'm also not too thrilled about it...just want to make life easier for people going back and forth between Clojure implementations. (I'm also not thrilled about `map` not creating a map, when `vector` creates a vector, `set` creates a set, etc...but that's life in the world of functional/Lisp languages, where `map` is a common verb.)

candid82:
Well, renaming `Int` to `Integer` would not be good since Java's (and Clojure's) `Integer` is a wrapper around `int`, which is 32-bit, whereas Joker's `Int` is a wrapper around Go's `int`, which is 64-bit on 64 bit platforms, so it's really equivalent to `Long`, not `Integer`. Perhaps the types should have been named `Long` and `Boolean`, but I don't really want to change the names now, unless it's causing real problems. One option could be adding `Long` and `Boolean` as aliases for `Int` and `Bool`, but this could potentially create confusion...

jcburley:
Aha, great point about `Int`! And worth documenting.

I think `Long` should be done only if it becomes clear that people really need that data type in real-life Joker code.

Presumably if someone really wants to pull in generic Clojure code using `Integer` (as well as `Boolean`, I guess), they can define their own aliases.

But `Bool` -> `Boolean` still feels like a win to me, though as (still) a Clojure-ecosystem newbie, I'm not strongly opined towards it at this time.

[...]

PR #244, opened 2019-08-05
Quick fix for 32-bit hosts

jcburley:
Possible fix for https://github.com/candid82/joker/issues/243.

### DIFF
--- a/core/object.go
+++ b/core/object.go
@@ -1014,6 +1014,10 @@ func (rat *Ratio) Compare(other Object) int {
 	return CompareNumbers(rat, AssertNumber(other, "Cannot compare Ratio and "+other.GetType().ToString(false)))
 }
 
+func MakeBigInt(bi int64) *BigInt {
+	return &BigInt{b: *big.NewInt(bi)}
+}
+
 func (bi *BigInt) ToString(escape bool) string {
 	return bi.b.String() + "N"
 }
--- a/std/time.joke
+++ b/std/time.joke
@@ -151,15 +151,15 @@
 (def
   ^{:doc "Number of nanoseconds in 1 minute"
     :added "1.0"
-    :tag Int
-    :go "int(time.Minute)"}
+    :tag BigInt
+    :go "int64(time.Minute)"}
   minute)
 
 (def
   ^{:doc "Number of nanoseconds in 1 hour"
     :added "1.0"
-    :tag Int
-    :go "int(time.Hour)"}
+    :tag BigInt
+    :go "int64(time.Hour)"}
   hour)
 
 (def
--- a/std/time/a_time.go
+++ b/std/time/a_time.go
@@ -10,11 +10,11 @@ import (
 var timeNamespace = GLOBAL_ENV.EnsureNamespace(MakeSymbol("joker.time"))
 
 var ansi_c_ = MakeString(time.ANSIC)
-var hour_ = MakeInt(int(time.Hour))
+var hour_ = MakeBigInt(int64(time.Hour))
 var kitchen_ = MakeString(time.Kitchen)
 var microsecond_ = MakeInt(int(time.Microsecond))
 var millisecond_ = MakeInt(int(time.Millisecond))
-var minute_ = MakeInt(int(time.Minute))
+var minute_ = MakeBigInt(int64(time.Minute))
 var nanosecond_ = MakeInt(int(time.Nanosecond))
 var rfc1123_ = MakeString(time.RFC1123)
 var rfc1123_z_ = MakeString(time.RFC1123Z)


rcarmo:
Thanks, it appears to have worked:

```bash
pi@quad ~/Library/Go/src/github.com/candid82/joker
 % git fetch origin pull/244/head:32bit-fix 
remote: Enumerating objects: 8, done.
remote: Counting objects: 100% (8/8), done.
remote: Compressing objects: 100% (5/5), done.
remote: Total 8 (delta 2), reused 4 (delta 2), pack-reused 0
Unpacking objects: 100% (8/8), done.
From https://github.com/candid82/joker
 * [new ref]         refs/pull/244/head -> 32bit-fix
pi@quad ~/Library/Go/src/github.com/candid82/joker
 % git checkout 32bit-fix
Switched to branch '32bit-fix'
pi@quad ~/Library/Go/src/github.com/candid82/joker
 % ./run.sh --version
v0.12.6
```

I will test the binary a bit and file other issues should they arise (I've been looking at the source and think I can find my way around at a high level by now).

candid82:
This will do as a temporary fix until I get to migrating from `int` to `int64` as a base type for `Int`. Thanks a lot, @jcburley!

jcburley:
> This will do as a temporary fix until I get to migrating from `int` to `int64` as a base type for `Int`.

Ah, here's what this reminds me of:

https://github.com/candid82/joker/issues/174

Perhaps `Integer` should be introduced as a 32-bit integer on all platforms first (as this seems to be what it is in Clojure+JVM), so people who really want that can migrate to it. (Not sure Joker has the installed base to justify this as a _requirement_, but it could still be helpful for us to migrate various other things such as tests that are currently commented out and such?)

Then either `Int` could be redefined as `int64`, or an explicit `Int64` type could be introduced, or possibly both (if someday `Int` might mean "at least 64 bits", i.e. potentially an `int128`).

candid82:
I'd really like to have just one integer type. I do wonder if making `Int` a wrapper for `int64` will affect performance on 32-bit architectures, but then again, performance is an explicit non-goal of Joker (outside of startup time). Still need to look into `int` -> `int64` migration (have been preoccupied with other stuff later).

jcburley:
> I'd really like to have just one integer type. I do wonder if making `Int` a wrapper for `int64` will affect performance on 32-bit architectures, but then again, performance is an explicit non-goal of Joker (outside of startup time). Still need to look into `int` -> `int64` migration (have been preoccupied with other stuff later).

Given that Joker already has a `BigInt` type, I wonder whether having `Int` always be 64-bit is really worth the trouble (aside from the possible performance impact, which I suspect would be fairly minor)?

What are the use cases a 64-bit `Int` enables (on 32-bit systems, since it's already 64-bit on 64-bit systems)?

candid82:
Having `Int` the same size on all platforms makes things consistent and predictable and avoids issues like [this](https://github.com/candid82/joker/blob/master/core/numbers.go#L60). That code assume `Int` to be 64-bit, and I am pretty sure that is not the only place where such assumption is made.

jcburley:
> Having `Int` the same size on all platforms makes things consistent and predictable and avoids issues like [this](https://github.com/candid82/joker/blob/master/core/numbers.go#L60). That code assume `Int` to be 64-bit, and I am pretty sure that is not the only place where such assumption is made.

Well, in that case (and other cases we could probably find fairly easily), an easy "fix" could probably be made that returns `Int` if the result fits, `BigInt` otherwise.

I'm concerned that changing the semantics of `Int` at this point would be a lot more work than just fixing problems like that and living with the inconsistency.

From a wider perspective, I'm a bit torn on this issue, because I see advantages to going with the Clojure+JVM "run-anywhere promise", but I don't think Joker (being Go-based) could ever really fulfill that promise to anywhere near the same extent as Clojure+JVM does, since that isn't a design goal of Go.

In other words, yes, theoretically `Int` might offer Joker users (as well as Joker developers) some more stability as far as that goes; but it's likely they'd experience plenty of other inconsistencies deriving from being hosted in Go (most or all of which reflect Go's favoring of native performance over "compile once, run anywhere").

So the other viewpoint I hold is to think of Joker as a Clojure-like wrapper around the Go language and runtime, in which case having `Int` == `int` makes a lot of sense and can be easily explained. Even within the code base, that equivalency should be fairly straightforward to keep in mind.

I prefer the latter viewpoint especially in light of the fact that Clojure+JVM defines no `Int` type, but rather `Integer`, which is the JVM-promised (I assume?) "signed 64-bit everywhere" type you're looking for. If Joker had called it `Integer` in the beginning, there'd be more reason to change it to meet the expectations of Clojure coders.

Regardless of what you decide, I hope to be able to help make it happen.

If `Int` remains as-is, I can help find and fix places in the code where 32-bit architectures might run into problems, such as the one you identified.

And if you decide to add `Integer` as `int64`, I'd like to help with that; or, if you change `Int` to `int64`, I can help with that as well.

However, those latter two options (supporting an explicit `int64` type) will also require me to do corresponding work on my `gostd` fork...which I've been looking forward to doing (as it seems strange to have so many `int64` values in the Go std converted into `BigInt` due to the lack of a native Joker type for `int64`); but I'd have to budget my time between that and upstream work on Joker, of course. (I think I could do the `gostd` work in just a few days.)

candid82:
Thank you for such a thoughtful input and for offering help. I've tried converting `Int` to `int64`, but that felt a bit like yak shaving. After going back and forth on this, I've decided to keep `Int` as is. I've also [marked](https://github.com/candid82/joker/commit/3e97d42061e36bc7a0a7363059d3dfe46b3145d6) problematic places. Most of them are in `time` and `strconv` namespaces, and can be easily resolved by using `BigInt` instead of `Int`. Others are trickier to resolve without introducing additional range checks at runtime. At this point I am not even sure any of these issues are worth resolving at all. Perhaps it will be sufficient to document that on 32-bit architectures truncation is possible for large integers (that don't fit into 32-bit int). Frankly, I doubt anyone will ever complain about this, and we can always get back to it if it turns out to be a bigger issue than it currently seems.

Issue #250, opened 2019-08-23
run.sh Fails for Cross-platform Builds


jcburley:
This was mentioned in https://github.com/candid82/joker/issues/248#issue-482644818.

The problem is that `go generate` and `tools/sum256dir/` are run at build time, but with the target platform in effect (due to `$GOOS` and `$GOARCH` being defined).

candid82:
Hm... why would one want to run `run.sh` with cross compiling? Even if it compiles, would not it fail to start since it's build for a different platform? Or am I missing something?

candid82:
I consider `run.sh` a helper script for development. When releasing, I just run `GOOS=<os> GOARCH=<arch> go build` (actually, I just run `build-all.sh` that runs that command for three supported platforms).

jcburley:
LOL, it _is_ called `run.sh` (not `build.sh`) after all, so good point!

As I'm primarily a developer, I think of it as a build script (and usually invoke it via my own ad-hoc generic `build` tool).

I'll withdraw the PR and maybe just offer one for `README.md`.

PR #251, opened 2019-08-23
Use $GOHOSTOS and $GOHOSTARCH for build-time generation

jcburley:
Fixes https://github.com/candid82/joker/issues/250.

Note that I'm not sure this is a perfect fix, insofar as there's also a $GOARM env var that has no $GOHOSTARM equivalent that I can see (on my systems, which are all non-ARM). That might mean one cannot build on a particular ARM system while targeting a different ARM system.

If someone who has an ARM system up and running can confirm that $GOHOSTARM is defined in their environment, perhaps just adding GOARM="$GOHOSTARM" to the environments for host builds would fix that issue.

### DIFF
--- a/run.sh
+++ b/run.sh
@@ -15,13 +15,16 @@ if [ "$1" == "-v" ]; then
   ./joker -e '(print "\nLibraries available in this build:\n  ") (loaded-libs) (println)'
 fi
 
-SUM256="$(GOOS="$GOHOSTOS" GOARCH="$GOHOSTARCH" go run tools/sum256dir/main.go std)"
-(cd std; ../joker generate-std.joke 2> /dev/null)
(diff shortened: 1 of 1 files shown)


pmonks:
I can confirm that `$GOARM` is required for cross-platform builds on ARM.  [Some Go Lang docs for this env var are here.](https://github.com/golang/go/wiki/GoArm)

jcburley:
> I can confirm that `$GOARM` is required for cross-platform builds on ARM. [Some Go Lang docs for this env var are here.](https://github.com/golang/go/wiki/GoArm)

Can you determine whether `$GOHOSTARM` is defined on such systems? I don't want to break native builds on ARM by assuming it's always defined; but, if it is always defined, I don't want to introduce more complexity to this PR (`run.sh`) trying to figure out whether and to what `$GOARM` should be set for building native (vs targeted) components.

jcburley:
Closing per https://github.com/candid82/joker/issues/250#issuecomment-524655200

PR #256, opened 2019-09-02
Support (doc <namespace>)

jcburley:
Also print cleaner diagnostic for (doc :any-keyword).

This enables e.g.:

```
user=> (doc joker.core)
-------------------------
joker.core
  Core library of Joker.
nil
user=> (doc joker.math)
-------------------------
joker.math
  Provides basic constants and mathematical functions.
nil
user=>
```

### DIFF
--- a/core/data/repl.joke
+++ b/core/data/repl.joke
@@ -40,6 +40,9 @@
          :name name-symbol
          :special-form true))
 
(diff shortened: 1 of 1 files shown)


candid82:
Thanks!

PR #258, opened 2019-09-03
Try initializing namespaces lazily

jcburley:
This is something of a proof of concept, as it doesn't seem to speed up official Joker much that I can measure on the two machines I'm using while on vacation (and my main development machine, at home, is even less likely to show much of a difference).

But, applied to my `gostd` fork, the resulting Joker definitely starts up faster, though still not as quickly as official Joker. (I think there's room for further improvement, in terms of delaying `var`-time initialization until `Init()` is called. That might help things further.)

That suggests this PR, or something like it, could keep Joker startup times from getting much longer even if/as more namespaces and/or functions (vars) are added to its default library.

Note that, while trying different ideas out, I was disappointed to discover that Go currently does not offer much in the way of practical build-time initialization, beyond scalar constants. That is, arrays and maps, in particular, cannot be initialized at build time and thus be present, already-initialized, when the resulting executable starts up. I hope to see improvements in this situation someday soon, and might look into how feasible it is to do and possibly offer to help with them.

### DIFF
--- a/core/environment.go
+++ b/core/environment.go
@@ -2,6 +2,7 @@ package core
 
 import (
 	"bufio"
+	"fmt"
 	"io"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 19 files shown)


candid82:
This is a pretty neat trick. Thanks!

PR #259, opened 2019-09-03
Add failing instance? test case

nenadalm:
Hi. There should be no linting errors.

`ex-info` (https://cljs.github.io/api/cljs.core/#ex-info) creates instance of `ExceptionInfo` (https://cljs.github.io/api/cljs.core/#ExceptionInfo), so the code should be valid.


candid82:
@nenadalm Thank you for submitting the test case. I fixed the issue and added reduced test case here: https://github.com/candid82/joker/commit/a8c637a5af7109d4c8fc2567413b87b526d6aba5

PR #261, opened 2019-09-09
Merge more startup-time laziness/deferral from gostd fork

jcburley:
This defers most everything in the `std/` libraries.

The one remaining item (that I know of) is `client` in `std/http/http_native.go`, which the `gostd` fork defers using a general mechanism that could be ported over to here as well, if deemed worthwhile.

### DIFF
--- a/std/base64/a_base64.go
+++ b/std/base64/a_base64.go
@@ -10,7 +10,9 @@ var base64Namespace = GLOBAL_ENV.EnsureNamespace(MakeSymbol("joker.base64"))
 
 
 
-var decode_string_ Proc = func(_args []Object) Object {
+var decode_string_ Proc
+
+func __decode_string_(_args []Object) Object {
 	_c := len(_args)
 	switch {
 	case _c == 1:
@@ -24,7 +26,9 @@ var decode_string_ Proc = func(_args []Object) Object {
 	return NIL
 }
 
-var encode_string_ Proc = func(_args []Object) Object {
+var encode_string_ Proc
+
+func __encode_string_(_args []Object) Object {
 	_c := len(_args)
 	switch {
 	case _c == 1:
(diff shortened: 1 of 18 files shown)


candid82:
Thanks! It probably doesn't make much different here, but I wonder how much this technique speeds up gostd fork?

candid82:
Also, what do you think can be deferred in `http_native.go`? I don't think deferring [client initialization](https://github.com/candid82/joker/blob/master/std/http/http_native.go#L15) would make any difference...

jcburley:
> Thanks! It probably doesn't make much different here, but I wonder how much this technique speeds up gostd fork?

It speeds it up just enough to notice, but it's still hard to measure.

But this, in combo with the original lazy-init stuff, seems to bring the `gostd` fork to within an "acceptable" range of startup delay, though by "acceptable" I mean I can see as much as 19% on one machine, 10% on another, but as little as basically 0% on a 3rd.

As I commented on the `gostd` Evaluation PR, I think the remaining time might be somewhat due to init costs for all those Go packages being added -- but haven't investigated to find out whether that's actually true (i.e. whether they actually do enough for it to matter).

jcburley:
> Also, what do you think can be deferred in `http_native.go`? I don't think deferring [client initialization](https://github.com/candid82/joker/blob/master/std/http/http_native.go#L15) would make any difference...

That's the only thing I've found in all those `std` libraries. And, no, I don't think it'd make any noticeable difference -- this entire PR doesn't seem to provide a measurable improvement AFAICT.

It's possible that doing perf measurements on some "slow" machines (say, 32-bit ARM or something?) might give us better numbers.

Anyway, this was all low-enough-hanging fruit that it seemed worthwhile, and on the whole it does seem to have eliminated a substantial amount (maybe a third or more?) of the excess startup cost of `gostd` versus official Joker.

jcburley:
> this entire PR doesn't seem to provide a measurable improvement AFAICT.

(I meant on the official (this version of) Joker.)

PR #262, opened 2019-09-10
Support joker.repl/doc of types

jcburley:
Not sure this is the best approach, but wanted to try it before working on adding (autogenerated) documentation of (Joker-wrapped) Go types in the `gostd` fork, to prevent the fork from straying further from the official version.

So, feedback highly welcome!

(Note that I added actual documentation to only the `Int` type, just to try it out.)

### DIFF
--- a/core/data/core.joke
+++ b/core/data/core.joke
@@ -3582,11 +3582,6 @@
   (^Map [^Associative m k ^Callable f x y z & more]
    (assoc m k (apply f (get m k) x y z more))))
 
-(defmacro type?
-  "Returns whether the object is a type"
-  {:added "1.0"}
-  ^Boolean [x] `(= (type ~x) ~Type))
-
 (defn coll?
   "Returns true if x implements Collection"
   {:added "1.0"}
--- a/core/data/core.joke
+++ b/core/data/core.joke
@@ -3582,6 +3582,11 @@
   (^Map [^Associative m k ^Callable f x y z & more]
    (assoc m k (apply f (get m k) x y z more))))
 
+(defmacro type?
+  "Returns whether the object is a type"
+  {:added "1.0"}
+  ^Boolean [x] `(= (type ~x) ~Type))
+
 (defn coll?
   "Returns true if x implements Collection"
   {:added "1.0"}
--- a/core/object.go
+++ b/core/object.go
@@ -305,7 +305,7 @@ func RegRefType(name string, inst interface{}, doc string) *Type {
 	if doc != "" {
 		doc = "\n  " + doc
 	}
-	meta := MakeMeta(nil, "(Concrete reference 'Type')"+doc, "1.0")
+	meta := MakeMeta(nil, "(Concrete reference type)"+doc, "1.0")
 	meta.Add(KEYWORDS.name, MakeString(name))
 	t := &Type{MetaHolder{meta}, name, reflect.TypeOf(inst)}
 	TYPES[STRINGS.Intern(name)] = t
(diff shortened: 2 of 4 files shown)


core/data/core.joke
```diff
@@ -3582,6 +3582,11 @@
   (^Map [^Associative m k ^Callable f x y z & more]
    (assoc m k (apply f (get m k) x y z more))))
 
+(defmacro type?
```
candid82: I think this macro is not very useful by itself and it's better not to expand core's surface area (it's already huge) and diverge from Clojure without necessity. So I think it probably makes sense to inline this into `doc` macro.

core/object.go
```diff
@@ -300,92 +301,108 @@ type (
...
 	return t
 }
 
-func regInterface(name string, inst interface{}) *Type {
-	t := &Type{name: name, reflectType: reflect.TypeOf(inst).Elem()}
+func regInterface(name string, inst interface{}, doc string) *Type {
+	if doc != "" {
+		doc = "\n  " + doc
+	}
+	meta := MakeMeta(nil, "(Abstract type, aka 'interface')"+doc, "1.0")
```
candid82: `aka` feels a bit informal. I think just `(Interface type)` would be better.

core/object.go
```diff
@@ -300,92 +301,108 @@ type (
 var TYPES = map[*string]*Type{}
 var TYPE Types
 
-func RegRefType(name string, inst interface{}) *Type {
-	t := &Type{name: name, reflectType: reflect.TypeOf(inst)}
+func RegRefType(name string, inst interface{}, doc string) *Type {
+	if doc != "" {
+		doc = "\n  " + doc
+	}
+	meta := MakeMeta(nil, "(Concrete reference 'Type')"+doc, "1.0")
```
candid82: Not sure I like quoting here. How about just `"(Concrete reference type)"` (and `"(Concrete type)"` below)?

core/procs.go
```diff
@@ -152,6 +152,11 @@ var procMeta Proc = func(args []Object) Object {
 		if meta != nil {
 			return meta
 		}
+	case *Type:
```
candid82: This should not be necessary. Embedding `MetaHolder` into `Type` should make it implement `Meta`, or am I missing something?

jcburley: This one has me stumped too: if I remove the `case *Type` entirely, `(doc Int)` and such stop working.

And if I make the first one into `case Meta, *Type:`, I get this from the compiler:

```
./procs.go:151:14: obj.GetMeta undefined (type Object has no field or method GetMeta)
```

I think the combining of a non-reference with a reference type match in a single `case` means the compiler can't assume anything further about the object (`obj` in this case) than that it's of the original declared type (`Object`, based on `args []Object` and assigning it from `args[0]`).

Does any other (useful) type of object, provided as an argument to `(meta ...)`, get passed as a reference, as `Type` does (in that `ProcMeta()` gets an `obj` of `*Type`)?

jcburley:
Thanks for the great feedback. I've made all the changes except for the unresolved conversation.

core/procs.go
jcburley: Re-reading that comment of mine, it's poorly worded, but you might figure it out anyway. Let me know if not....

candid82: Hm... I'll look into this issue later, but the PR looks great as is. Merging. Thanks!

candid82: Just looked into this more. `Meta` interface has two methods: `GetMeta` and `WithMeta`. `MetaHolder` only implements `GetMeta`, so embedding it is not enough to implement `Meta`. In case of `Type` it doesn't even make sense for `Type` to implement `Meta`, because calling `WithMeta` on a type is nonsense (meta data is read-only for types). So I think your implementation is right.

PR #263, opened 2019-09-10
Quote namespace doc string

jcburley:
This should have no effect on official Joker, but (besides being
correct and consistent with how the :doc tags are treated elsewhere)
it fixed the `gostd` fork when reading the generated online docs, on
my website, for `html/template`, which include this gem:

```
"<script>alert('you have been pwned')</script>"
```

(I was actually getting that alert whenever reloading that page!)


candid82:
Thanks!

PR #264, opened 2019-09-10
Show unexpected errors from the generate-std.joke step

jcburley:
The current behavior is to silently fail, which is confusing (when it happens to us developers).

However, this approach relies on failures printing an unexpected diagnostic; merely exiting with a nonzero return code no longer suffices.

I'm not sure whether it's worth fixing that possibility, given that a better fix might be to inhibit the expected warnings (and blank lines) entirely, rely solely on the return code, and not /dev/null the output.

### DIFF
--- a/run.sh
+++ b/run.sh
@@ -16,7 +16,12 @@ if [ "$1" == "-v" ]; then
 fi
 
 SUM256="$(go run tools/sum256dir/main.go std)"
-(cd std; ../joker generate-std.joke 2> /dev/null)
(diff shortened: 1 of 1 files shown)


candid82:
Yeah, this feels a bit hacky, but will do for now. I have not run into issues with `generate-std.joke` yet, so I am not in a good position to evaluate different solutions here.
Thanks!

jcburley:
Yes, definitely hacky. Maybe someday a way for `generate-std.joke` script to mark things as "expected", so they're not warned about, will allow removal of this and the `/dev/null`-ing of its output.

(I've often 'broken" `generate-std.joke` in my Joker work, so this definitely helps me....)

PR #265, opened 2019-09-11
Add joker.private namespace with (types) to return map of registered types

jcburley:
E.g.:

```
user=> (doseq [[k v] (joker.private/types)] (println k))
Comparable
Gettable
Ref
MapSet
[...]
```

What do you think? Might be handy; and I'm thinking a `joker.private` namespace might be a useful thing generally.

Note that I've got this working in my private `gostd` fork (not yet pushed to my public repo):

```
user=> (doseq [[k v] (joker.private/types)] (doc (identity v)))
-------------------------
Counted
  (Abstract type, aka 'interface')
-------------------------
Nil
  (Concrete 'Type')
-------------------------
Regex
  (Concrete 'Type')
-------------------------
Set
  (Abstract type, aka 'interface')
[...]
```

(I'm working on refactoring how `gostd` handles Go variables, objects, and types, so things like this are helpful to me for use as well as for understanding how Joker, and Clojure generally, view types.)


jcburley:
One of my goals here is autogeneration of documentation of Joker types, along with things residing in Joker namespaces.

candid82:
Instead of introducing new namespace, do you think it makes more sense to make `ns-map` function return types in addition to vars? This is how Clojure works, so doing so would have the additional benefit of bringing Joker's behavior closer to Clojure.

jcburley:
> Instead of introducing new namespace, do you think it makes more sense to make `ns-map` function return types in addition to vars? This is how Clojure works, so doing so would have the additional benefit of bringing Joker's behavior closer to Clojure.

That's a much better idea, thanks! Will update the PR accordingly.

(I do have some ideas for Joker that might include stuff best put in a `joker.private` namespace, but if there's no need for it now, best to put that off. Also, I originally called it `joker.internal`, but Go refuses to compile a package named `internal`.)

jcburley:
> > Instead of introducing new namespace, do you think it makes more sense to make `ns-map` function return types in addition to vars? This is how Clojure works, so doing so would have the additional benefit of bringing Joker's behavior closer to Clojure.
> 
> That's a much better idea, thanks! Will update the PR accordingly.

Checking Clojure's behavior, it seems the types are actually added into each namespace, so `ns-map` doesn't actually do anything "special" to return them, so e.g. `ns-unmap`  works on them:

```
user=> (get *ns* 'Integer)
java.lang.Integer
user=> (ns-unmap *ns* 'Integer)
nil
user=> Integer
CompilerException java.lang.RuntimeException: Unable to resolve symbol: Integer in this context, compiling:(NO_SOURCE_PATH:0:0) 
user=> (get (ns-map *ns*) 'Integer)
nil
user=> (get (ns-map 'clojure.core) 'Integer)
java.lang.Integer
user=>
```

I think that's a sufficiently different PR to do (which I want to think about more first, unless you want to take care of it) that I'll just Close this one out after having created https://github.com/candid82/joker/issues/267 to track this.

jcburley:
Reopening to go forward per https://github.com/candid82/joker/issues/267#issuecomment-531633047 (and retitled this PR accordingly).

jcburley:
Makes more sense to push a new PR after all.

Issue #267, opened 2019-09-16
ns-map Doesn't Return Types


jcburley:
In Clojure, this works (from the REPL, for example):

```
user=> (get (ns-map *ns*) 'Boolean)
java.lang.Boolean
user=>
```

But Joker doesn't add its builtin types to new namespaces:

```
user=> (get (ns-map *ns*) 'Boolean)
nil
user=>
```

Since Clojure adds types to even a brand-new namespace created by e.g. `(in-ns 'mostly-empty)`, Joker should probably do the same thing.

jcburley:
So my first quick cut at this runs into the problem that `Namespace.mappings` maps `*string` to `*Var`, but I want to add types (of type`*Type`) to the mappings....

It seems this won't be a really trivial thing to support (though maybe not particularly difficult either). Doubt I'll look into it further tonight.

candid82:
So I gave it more thought and I am now not sure it's a good idea to replicate Clojure's behavior here. As of now, Joker's set of types is global and fixed (immutable), so it doesn't feel natural to me to have them in a namespace's mappings... Perhaps your [original PR](https://github.com/candid82/joker/pull/265) is a simpler way to expose the types. I am not sure about putting `types` function in `joker.private` namespace though. What I've done in the past is to put `^:private` functions with `__` postfix in `joker.core`, i.e. https://github.com/candid82/joker/blob/master/core/data/core.joke#L2891 I think I'll be OK with doing `(defn ^:private types__ ...)` in `joker.core`.

jcburley:
That would be a lot simpler to implement!

jcburley:
Workaround available as https://github.com/candid82/joker/pull/268 (assuming it's accepted); meanwhile, closing as not likely to implement this unless and until types are handled rather differently in Joker (more like Clojure, in that they'd be namespace-hosted instead of "omnipresent", I suppose).

PR #268, opened 2019-09-16
Add joker.core/types__ as private function

jcburley:
Alternate (simpler) approach to addressing https://github.com/candid82/joker/issues/267 per discussion.


candid82:
👍

PR #269, opened 2019-09-16
Doc all types

jcburley:
Assuming https://github.com/candid82/joker/pull/268 is merged, this could then be merged to actually use the new function when generating docs.

(I don't have strong feelings about the stylistic changes I made to the `index.html` page to accommodate listing types; just wanted to get it working and look fairly reasonable without being too intrusive.)


candid82:
Thanks!

PR #270, opened 2019-09-16
Use first line only in namespace summaries, and HTML-escape :doc strings

jcburley:
This should improve consistency and readability of the generated docs, and avoid issues when "special" character (in HTML) are in `:doc` strings.

### DIFF
--- a/core/data/walk.joke
+++ b/core/data/walk.joke
@@ -20,11 +20,11 @@
 
 (ns
   ^{:author "Stuart Sierra",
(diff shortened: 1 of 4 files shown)


jcburley:
This might conflict with https://github.com/candid82/joker/pull/269, which adds another case where escaping is needed (on the `gostd` branch anyway), so maybe merge that first?

candid82:
Thanks! I really appreciate all these documentation improvements!

PR #274, opened 2019-10-03
New --exit-to-repl [socket] and --file options

jcburley:
Supports dropping into REPL (optionally socket REPL), instead of simply exiting, after successful `--eval` or `--file`.


candid82:
Thanks!

PR #281, opened 2019-10-18
Update README.md

prochac:
fix URL to Arch AUR package


prochac:
haawda commented on 2018-12-01 12:48
This should be named joker-bin, as it does not build from source.

https://aur.archlinux.org/packages/joker-bin#comment-672666

candid82:
Thanks!

PR #283, opened 2019-10-22
Save --file/--expr input for history playback when --*-to-repl in effect

jcburley:
This is very handy for retrying (possibly with editing), in the dropped-into-repl, previous lines from a `--file` (or even `--expr`).

E.g. after `joker --exit-to-repl --file <somefile>.joke` and finding oneself in the repl, at which point `<somefile>.joke` will have been read, an up-arrow will (via `readline`) bring up the most recent line read from `<somefile>.joke`, as the history of those reads will have been saved.

Shouldn't cost anything (measurable) unless `--*-to-repl` has been specified (and of course `--file`/`--expr` is in effect).

Note that the extra "logic" (of initializing the new var to `true` and including it in the conditional) is currently unnecessary, but allows dropping in another option to disable just this one feature, if one wishes. I don't know why anybody would want that (without also wanting `--no-readline`, which will have the same effect), though.

Might be nice to someday also support populating `*1`, `*2`, `*3`, and `*e`.

### DIFF
--- a/main.go
+++ b/main.go
@@ -34,6 +34,25 @@ import (
 	"github.com/pkg/profile"
 )
 
+var dataRead = []rune{}
+var saveForRepl = true
+
+type replayable struct {
+	reader *Reader
+}
+
+func (r *replayable) ReadRune() (ch rune, size int, err error) {
+	ch = r.reader.Get()
+	if ch == EOF {
+		err = io.EOF
+		size = 0
+	} else {
+		dataRead = append(dataRead, ch)
+		size = 1
+	}
+	return
+}
+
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -34,6 +34,25 @@ import (
 	"github.com/pkg/profile"
 )
 
+var dataRead = []rune{}
+var saveForRepl = true
```
candid82: Where is this option read?

candid82: Never mind, I should have read the description more carefully.

candid82:
Thanks!

PR #285, opened 2019-10-23
Add other meta tags

jcburley:
This treats the `std/*.joke` more like the `core/data/*.joke` files in that "all" the metadata on each public symbol is provided during interning.

That being said, for now this PR doesn't really handle anything (for values of meta tags) other than strings and integers. Any other type is stringized.

While this is theoretically useful in canonical Joker, I found this helpful (for generating better documentation) in the `gostd` fork.


candid82:
Thanks!

PR #286, opened 2019-10-31
Plan9 support

majiru:
Some function modifications to allow joker to be built for plan9.


candid82:
@majiru Thanks! I'll merge since this looks simple enough, but plan9 won't be officially supported any time soon (if ever).

PR #288, opened 2019-11-03
Arg Literals Followed By Commas

youngnh:
Noticed that arg literals in anonymous functions throw a parsing error if followed by a comma:

```
$ echo "(#(prn {:a %, :b 2}) :c)" | joker --lint -
<stdin>:1:16 Read error: Arg literal must be %, %& or %integer
```

while clojure will parse and run it just fine:

```
$ echo "(#(prn {:a %, :b 2}) :c)" | clj -
{:a :c, :b 2}
```


youngnh:
hmm, I just realized that this isn't just an issue with `--lint` maybe I should move the tests...?

jcburley:
LGTM, though I think `'%'` should also be excluded from the `case` statement that returns `true` in `isTerminatingMacro()` (which "fixes" `#(+ %%)` being accepted as valid Clojure, which 1.9.0 doesn't accept).

I might submit that as a distinct PR, after this one is (presumably) merged, adding an appropriate test.

candid82:
@youngnh Hey Nate, thanks for the fixes (including readme fixes)! I don't think it matters that much where the test is located as long as there *is* one.
@jcburley you are right about `%%` case! I'll let you PR the fix since you found it, unless you want me to fix it.

PR #289, opened 2019-11-04
Disallow arg literal of arg literal

jcburley:
This "reserves" `%%`, as Clojure 1.9.0 does, instead of treating it as `% %` (i.e. with a space in between).

There's also a slight code cleanup, in that `isWhitespace()` is used by `isDelimiter()` instead of that having its own version of the former's code.

And a test.


candid82:
Thanks!

PR #290, opened 2019-11-07
Add more constants and functions from Go math package

jcburley:
Might do more later, but had some downtime and decided these might be useful for now.


jcburley:
Note that I made some guesses as to appropriate (Clojure-style) naming of things. Am not sure they're entirely proper....

candid82:
Wow, this is great! Thank you very much, @jcburley!

PR #292, opened 2019-11-10
More docs on adding namespaces to Joker executable

jcburley:
In light of recent discussion on Slack, and because I'm still learning (and sometimes have to re-learn) how this all works, here's some basic info that I hope is correct.

(It needs to be expanded upon at some point. E.g. what is the format of `:go` metadata?)


candid82:
This is extremely useful! I admit I myself sometimes forget how things work. Thanks a lot, @jcburley !

PR #293, opened 2019-11-11
Generate code for core before depending on it for further generation

jcburley:
gen_data.go imports Joker's core, which requires that to compile cleanly.

But gen_types.go generates code that is built into Joker's core, while not depending upon it.

Therefore, the "go:generate" lines that run gen_types.go should come before the one that runs gen_data.go.

Test this by editing (say) core/types_assert_gen.go to make it uncompilable and rebuilding (via run.sh or similar).

The "go generate ./..." step should regenerate core/types_assert_gen.go, replacing the uncompilable version with a working version, before gen_data.go is then run and imports core.


candid82:
Thanks!

PR #294, opened 2019-11-11
Add joker.markup namespace

hlship:
Patterned on Hiccup, this namespace allows for rendering of XHTML (and variants) markup from vectors and strings.

It is not a direct port of Hiccup, as it is strictly interpretive, but its basic API is compatible with Hiccup (nearly all tests were just copied over unchanged), and should be quite sufficient for generating blog posts, XML configuration files, or other similar operations in Joker's intended domain.

### DIFF
--- a/core/data/markup.joke
+++ b/core/data/markup.joke
@@ -1,0 +1,166 @@
+(ns joker.markup
+  "Renders HTML, XML, or XHTML markup to a string, based on prior work in Hiccup.
+
+  https://github.com/weavejester/hiccup"
+  {:added "0.14.0"}
+  (:require
+   [joker.html :refer [escape]]
+   [joker.string :as str]))
+
+(defn ^:private named?
+  [v]
+  (or (keyword? v)
+      (string? v)
+      (symbol? v)))
+
+(defn ^:private as-str
+  [v]
+  (cond
+    (named? v)
+    (name v)
+
+    (nil? v)
+    ""
+
+    (number? v)
+    (str v)
+
+    :else
+    (str v)))
+
+(defn ^:private xml-mode?
+  [mode]
+  (#{:xml :xhtml} mode))
+
+(defn ^:private html-mode?
+  [mode]
+  (#{:html :xhtml} mode))
+
+(defn ^:private end-tag
+  [mode]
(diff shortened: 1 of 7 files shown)


didibus:
I personally find the namespace name a bit misleading. Since hiccup isn't a markup language on it's own. I would intuitively expect markup to be more a utils namespace related to multiple markup languages like markdown, org, mediawiki, creole, etc.

I'd like to propose we start a new namespace group such as:

* joker.templating.html

or

* joker.template.html

In the future, if joker adds support for templating to other outputs this structure would naturally extend to them.

I'm also okay with keeping it flat if that's prefered and just call it:

* joker.html-template

Or

* joker.html-templating

And I also wouldn't be opposed if @weavejester blesses it, that we treat Hiccup as the name of the syntax and template language and not the specific implementing lib of it in Clojure. In that way the namespace could simply be called:

* joker.hiccup

Or

* joker.hiccup-templating

Or

* joker.template.hiccup

Or you know, any combination of that sort.

didibus:
Also, I was wondering if this follows Hiccup 1 or Hiccup 2 syntax?

The difference is that Hiccup 2 escapes by default, and is thus more secure. And requires you to be explicit when you don't want escaping. While Hiccup 1 doesn't escape by default.

I think it be nicer going with Hiccup 2, for it's better security default. But at the same time, the use cases for joker might not make use of dynamic Hiccup generation from untrusted user input as often as it does in Clojure. And Hiccup 2 isn't as popular in the Clojure world, as it came out later and is still it seems in a pre-release phase.

weavejester:
I'm fine with you using the name "hiccup" if you want.

hlship:
It follows hiccup2 syntax, escaping by default;  there's a `raw-string` function to wrap a string that is safe to insert as-is.

hlship:
let's see what Roman thinks about package naming then.

candid82:
@hlship Thank you for putting this together! I think `joker.hiccup` would be a clearer name. Also, `:added` meta is currently set to `1.0` for everything (which is not useful until v1.0 is actually reached, but oh well). I'll merge and make the changes. Thanks a lot!

PR #297, opened 2019-11-22
Add joker.os/remove{,-all}

jcburley:
Per https://clojurians.slack.com/archives/C9VURUUNL/p1574457045102100, seems like these should be available without resorting to the `gostd` fork at this time.


candid82:
Thanks!

PR #298, opened 2019-11-25
Add missing meta tag

jcburley:


candid82:
Thanks! 
Btw, I'd be interested in your thoughts on the concurrency implementation that this `go` macro represents. This is very experimental at the moment, and I will probably introduce channels and make goroutines return channels (like core.async) rather than futures. I will also delete `*-async` functions as goroutines are a more general way of achieving concurrency (that is, `(go (joker.http/send...))` instead of `(joker.http/send-async)`). 
This "cooperative multitasking" (a goroutine has to do I/O or deref a Future to allow other goroutines to run) seems to work for the use cases I have in mind for Joker, but it may be a bit confusing (a goroutine may never run if the root goroutine doesn't do I/O or derefs a Future).
Let me know if you have any thoughts on this.
Thanks again!

candid82:
FWIW, I've seen the same concurrency model in Python's [gevent](http://www.gevent.org/intro.html#cooperative-multitasking).

PR #300, opened 2019-11-30
Add joker.tools.multi

hlship:
This PR adds namespace joker.tools.multi; multi allows a single Joker script to contain any number of individual commands, each with its own options and (positional) arguments.

- commands are identified via metadata within specific namespaces
- adds a "help" command, and a "-h/--help" option to each command
- Introduces facilities similar to joker.tools.cli to discuss how to parse, validate, and collect command line arguments

Note: have not updated Joker documentation with this PR.


candid82:
This is an interesting feature that may come handy in certain scenarios, but it doesn't feel like it should be built into Joker. Unlike `tools.cli`, which is fairly low level API that is already widely adopted by the Clojure community, `tools.mutli` seems to be pretty high level and opinionated, and it has not been battle tested yet. It also addresses relatively narrow set of use cases.
I think it's better to keep it as a separate library. Luckily, Joker allows you to easily include external dependency with `ns-sources` (thanks to @charlesg3!). For example, this works:
```clojure
#!/usr/bin/env joker

(ns-sources
 {"joker.tools.multi" {:url "https://raw.githubusercontent.com/candid82/joker/045af6f5d514d9f374c4908842b8996e46d7b8be/core/data/tools_multi.joke"}})

(ns my.test
  (:require [joker.tools.multi :as multi]))

(defn ^{:command true
        :command-opts [["-v" "--verbose" "Enable verbose logging"]]
        :command-args [["HOST" "System configuration URL"
                        :validate [#(re-matches #"https?://.+" %) "must be a URL"]]
                       ["DATA" "Data to configure as KEY=VALUE"
                        :id :key-values
                        :parse-fn (fn [s]
                                    (when-let [[_ k v] (re-matches #"(.+)=(.+)" s)]
                                      [(keyword k) v]))
                        :update-fn (fn [m [k v]]
                                     (assoc m k v))
                        :repeatable true]]}
  configure
  "Configures the system with keys and values"
  [parsed-opts]
  (pprint (select-keys parsed-opts [:options :arguments])))

(multi/dispatch {:tool-name "test" :namespaces ['my.test] :arguments *command-line-args*})
```

`ns-sources` will download the file from the server and cache it in `~/.jokerd/deps` directory. Please note that I had to [tweak](https://github.com/candid82/joker/commit/2f3259a79358be6a9d8050a891ac3e627491b2e8) the loading code a bit to support direct file urls, so if you want to try the above example use the latest master.

hlship:
I was not aware of `ns-sources` before; this is great news, much easier to maintain little libraries like this outside of Joker itself.

PR #302, opened 2019-12-18
Add joker.repl to *loaded-libs*

jcburley:
Happened to notice this and can't see any reason for it to be missing!


candid82:
Thanks!

PR #303, opened 2019-12-20
Create joker.better-cond from Engleberg/better-cond

hlship:
This PR adds namespace `joker.better-cond`, based on code extracted from https://github.com/Engelberg/better-cond.

I've found that nearly every Joker script I've written has needed `cond-let` (a more limited version of better-cond's version of `cond`).

Original work licensed under Eclipse Public License.


candid82:
Thanks!

PR #304, opened 2019-12-21
Fix deps.joke to use source filename directly

jcburley:
Instead of guessing which argument, in the original invocation of Joker,
has the filename, or even using the Joker-provided `*command-line-args*`
variable, just use `*main-file*` directly (it's a new-ish feature).

This allows the test to pass even if Joker options (not options to the
Joker script being invoked) are provided. E.g. this works:

```
./joker --evaluate tests/run-eval-tests.joke
```

### DIFF
--- a/tests/eval/deps.joke
+++ b/tests/eval/deps.joke
@@ -3,8 +3,7 @@
             [joker.string :as str]))
 
 (def lib-dir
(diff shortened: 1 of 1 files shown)


candid82:
Thanks!

PR #305, opened 2019-12-21
Improve tty testing so it works outside of the eval-tests.sh script

jcburley:
This allows `./joker tests/run-eval-tests.joke` (run from a command line,
so stdin is a tty) to support tty-related testing.


candid82:
Thanks!

PR #307, opened 2019-12-26
Stabilize most hash values

jcburley:
E.g. `(hash :hey)` will now return the same value across invocations, builds, and OSes. Ditto `(hash "hey")`, `(hash 'hey)`, and others.

This makes Joker behave more like Clojure.

Not all objects are hereby given stable hashes; they'll still differ from run to run, as they're pointer-based.

However, this should cover all the objects created during core namespace initialization, which should improve the ability for startup-time optimization to generate static (file-scope) initialization code (in Go), rather than having to rely on runtime initialization (after pointer values are established).


core/environment.go
```diff
@@ -114,9 +117,9 @@ func NewEnv(currentNs Symbol, stdin io.Reader, stdout io.Writer, stderr io.Write
 }
 
 func (env *Env) SetStdIO(stdin io.Reader, stdout io.Writer, stderr io.Writer) {
-	env.stdin.Value = &BufferedReader{bufio.NewReader(stdin)}
-	env.stdout.Value = &IOWriter{stdout}
-	env.stderr.Value = &IOWriter{stderr}
+	env.stdin.Value = MakeBufferedReader(stdin, stdinHashValue)
```
candid82: So stdin, stdout and stderr will always have the same hash values regardless of the actual objects?

core/ns.go
```diff
@@ -59,14 +59,17 @@ func (ns *Namespace) AlterMeta(fn *Fn, args []Object) Map {
...
 }
 
+const nsHashMask uint32 = 0x90569f6f
+
 func NewNamespace(sym Symbol) *Namespace {
 	return &Namespace{
 		Name:     sym,
 		mappings: make(map[*string]*Var),
 		aliases:  make(map[*string]*Namespace),
+		hash:     sym.Hash() ^ nsHashMask,
```
candid82: Is `^ nsHashMask` part there just to make the namespace's hash different from symbol's hash?

core/channel.go
```diff
@@ -36,15 +37,19 @@ func (ch *Channel) GetType() *Type {
...
+	return ch.hash
 }
 
 func (ch *Channel) WithInfo(info *ObjectInfo) Object {
 	return ch
 }
 
-func MakeChannel(ch chan FutureResult) *Channel {
-	return &Channel{ch: ch}
+func MakeChannel(ch chan FutureResult, hash uint32) *Channel {
```
candid82: Looks like all calls to `MakeChannel` pass `0` as `hash`, so this change doesn't seem meaningful to me. In general, here and in other places is seems odd to me that the hash value is passed as an argument rather than being derived from the "content" of an object itself. And mutable objects like channels and readers should have distinct hash values per instance anyways...

core/procs.go
```diff
@@ -1331,18 +1331,20 @@ var procArrayMap Proc = func(args []Object) Object {
 	return res
 }
 
+const bufferHashMask uint32 = 0x5ed19e84
+
 var procBuffer Proc = func(args []Object) Object {
 	if len(args) > 0 {
 		s := EnsureString(args, 0)
-		return &Buffer{bytes.NewBufferString(s.S)}
+		return MakeBuffer(bytes.NewBufferString(s.S), s.Hash()^bufferHashMask)
```
candid82: Since `Buffer` object are mutable, I don't think it's right to assign them a hash value based on their initial content...

core/environment.go
jcburley: Yes. At first this seemed strange, and I'm still on the fence about it, because:

- On the one hand, this is currently used only by the srepl() code, so there shouldn't be any conflict between this trio and what the repl() code would set up.
- Further, it might make sense (though I can't say why!) to have the same hash values for whatever comprises the "repl trio", regardless of what the repl uses for input.
- Yet, one could reasonably say that they're different file descriptors (which they are), so they should have different values.
- Clojure assigns them different values, but that makes sense since a Clojure sRepl is in addition to (instead of in place of) its standard repl. (In fact, it can handle multiple sRepls in a single instance). This might arguably be a good reason to keep the values the same for now, unless/until Joker also supports multiple active Repls per instance, so code could possibly detect that?

I can't think of any slam-dunk reasons to go with one or the other approach! Let me know whether you want me to use distinct hash values for these.

core/ns.go
jcburley: Yes. There's an existing such mask being added for `(Symbol)Hash()`. (Want me to convert that into a similar form as I'm using above, e.g. make a var for it and use `^` instead of `+`, for consistency?)

core/channel.go
jcburley: The only reason for the extra argument is to remind any future Joker maintainers, who might add new calls, to strive to compute a more-stable hash than would be assigned by default.

I'm happy to remove those extra args, though, because it's so unlikely they'll ever be used (in the cases where all current callers pass 0).

Note that I think it's a tad "safer" to compute the hash value for a new object once and store it, as this PR does for a few Object types, than to compute repeatedly. My thinking is that, since it's an "unsafe" computation (from the pointer to the object), the existence of a hash value won't prevent the GC from moving the object if needed to allocate more memory (not sure Go's GC ever does this -- it'd have to adjust all legit pointers to the object). By memoizing the value once and storing it in the object, the hash value for the object won't change even if the object itself moves. Would be a fun bug to track down, albeit highly unlikely to ever happen....

core/procs.go
jcburley: Indeed, I was being overzealous trying to find something "stable". Besides the fact that there's no reason to assume this would be stable (the initial content could easily differ from run to run), it'd result in different buffers having identical hashes merely because they started out with the same content (which could be the empty string, versus a nil argument).

core/environment.go
jcburley: Interestingly, while Joker assigns the same sRepl socket (from `l.Accept()`) to each of the three variables yet gives them differing hashes, Clojure gives (again, for an sRepl) `*in*` one hash while using the same hash for both `*out*` and `*err*`.

It's not clear to me what Joker should do here. Changing to be more like Clojure should probably be a distinct PR though.

jcburley:
I've updated the PR (and tested the results) per the Resolved discussions, above, but there remain a few open questions.

core/environment.go
candid82: I think we should keep hash semantics clear and consistent with equality unless we have reasons to do otherwise. Two different (not equal) objects should have different hashes.

candid82:
My only remaining objection is about using the same hashes for stdin/stderr/stdout. I think they should be different, and also hashes should not be passed as arguments to `Make*` functions (and rather should be calculated based on object's content).

core/environment.go
jcburley: That sounds good. I've updated the PR to better approximate what Clojure does, so the hash values for an sRepl are different from for a repl, still stable, and identical for `*out*` and `*err*`.

jcburley:
> My only remaining objection is about using the same hashes for stdin/stderr/stdout. I think they should be different, and also hashes should not be passed as arguments to `Make*` functions (and rather should be calculated based on object's content).

As mentioned above, I've changed the PR to use different hashes for sRepl versus Repl, so it's much more like Clojure.

I'm happy to change how the PR currently works for the two remaining such `Make*` functions, but how do you suggest they work, since Go doesn't expose the (concrete) contents of `io.Reader` and `io.Writer` objects that I can see?

That sort of thing is what led me down the path of having some `Make*` functions require the caller to provide a hash value in the first place.

At first, I looked into using reflection (and munging the "private/public" status of fields) to reach down into those objects and find -- say, the underlying file descriptor.

But they're abstract (`interface{}`), so there's no assurance of the ability to do so for any arbitrary type. Granted, Joker doesn't support arbitrary-type construction (Go doesn't at runtime); but it still seemed like a deep hole to go down.

Even if one could ultimately grab the underlying FD (0, 1, or 2 for the std items), I'm not sure that really gets at "content" any more than does just using a pointer (which is unstable, of course).

I.e. a pointer isn't "content". But neither is an FD, really; it's just whatever file or socket happens to be open on that particular FD at that particular time. Indeed, that can (in theory anyway, e.g. in a C program such as bash) change even for the std FDs -- whatever starts out on FD 0 for a process might no longer be open on FD 0 later when a new thing (like an `io.Reader`) is wrapped around it. In such a case, two objects wrapping FD 0 might "look" the same, but really aren't, from the perspective that one might well be "dead" (refers to a since-closed instance of FD 0); and they shouldn't compare the same.

(Similarly, for a network socket, the IP/port tuples for both sides of the connection aren't really "content" despite being "concrete". Nor is a URL "content"! They're all just pointers, or handles.)

By contrast, the callers of these `Make*` functions have some knowledge of the context that those functions can't divine from their (non-hash) arguments: whether the thing being made wraps an initial (upon-processing-invocation) std port, the first (only, for now) sRepl, or something else (and right now Joker doesn't seem to have any other callers, besides the one access solely via `line-seq`, which uses the pointer-as-hash approach).

Given all that, I think the PR currently represents a reasonable compromise, though versus what exactly I can't say (maybe it's "just right" now?). Though, as noted in the changed code for the sRepl case, it'll need changing if multiple sRepls are supported....

jcburley:
> By contrast, the callers of these `Make*` functions have some knowledge of the context that those functions can't divine from their (non-hash) arguments: whether the thing being made wraps an initial (upon-processing-invocation) std port, the first (only, for now) sRepl, or something else (and right now Joker doesn't seem to have any other callers, besides the one access solely via `line-seq`, which uses the pointer-as-hash approach).

Upon further reflection, since Clojure doesn't stabilize even the first sRepl std hashes, Joker shouldn't do that either. Simplifies the code a bit further as well. Pushed.

candid82:
I am not 100% sure what problem we are solving by making reader's and writer's hashes "stable". I don't think Clojure works that way either, e.g.:
```
> clj
Clojure 1.10.1
user=> (type *in*)
clojure.lang.LineNumberingPushbackReader
user=> (hash *in*)
148635643
user=>
/Users/candid/work/eReceipts-services/query-service (130)
> clj
Clojure 1.10.1
user=> (hash *in*)
1998949977
```

It seems to me that using pointer values should do just fine...

jcburley:
> I am not 100% sure what problem we are solving by making reader's and writer's hashes "stable". I don't think Clojure works that way either, e.g.:
> 
> ```
> > clj
> Clojure 1.10.1
> user=> (type *in*)
> clojure.lang.LineNumberingPushbackReader
> user=> (hash *in*)
> 148635643
> user=>
> /Users/candid/work/eReceipts-services/query-service (130)
> > clj
> Clojure 1.10.1
> user=> (hash *in*)
> 1998949977
> ```
> 
> It seems to me that using pointer values should do just fine...

Well, that's weird -- I see the same (mis?)behavior, in that doing `(type *in*)` first, before `(hash *in*)`, results in a different hash value! But it's often the same across multiple invocations when `(type *in*)` is done (first) as well as when it isn't. (This is true whether I invoke it in an `rlwrap` or just straight via `java`.)

Similarly, if I do `(type *in*)` _twice_ before `(hash *in*)`, I see yet another ("stable"!) hash value! I wonder what in the world is going on there? Maybe it doesn't initialize the hash until it first needs it, and whatever else it does first affects that? Haven't looked at that particular code.

I guess this means Clojure isn't striving to stabilize those particular hashes after all. I tried this with a symbol and a keyword, but those hashes remained stable even when I asked for them after asking for the types of them first.

That leaves only (AFAIK) the original reason I wanted to stabilize hashes generally: to make as much initialization "invariant" as possible, to improve startup speed versus leaving things "variant" (i.e. done at runtime during startup).

Since at this point we're talking about only three variables (`*in*`, `*out*`, and `*err*`), which I don't think get "embedded" into anything else in the way that, say, `Keyword` values do, I guess it's not a big deal. I'm sure my startup-speedup code can easily handle having these three variables being (partially) initialized at runtime, since it handles `*file*`, `*main-file*`, `*command-line-args*`, etc.

In fact, some things like that could even be lazily-initialized (`*in*` bound to `os.Stdin`), so they're not even done at runtime startup, saving a tiny bit more of time. E.g. not every invocation is going to look at `*main-file*`, `*command-line-args*`, etc. That would be entirely separate work from this PR or my speedup code, I would think; and I'm not sure it's worth doing, but it's interesting to think about in theory.

Let me know if you want me to back out that aspect of this PR.

Otherwise, I think it's good to go at this point...?

jcburley:
It did (finally!) occur to me to test `(hash 'foo)` then `(hash 'bar)` versus the opposite order in Clojure. The hashes remained stable vis-a-vis the symbol names. Ditto keywords. So at least _those_ aren't pointer-based.

candid82:
Yes, I think symbol's and keyword's hashes are value based while Java objects (including readers and writers) are pointer based. Joker currently calculates symbol's and keyword's hashes based on pointers because they are consistent with equality within single Joker invocation (that is, two equal symbols or keywords will have the same `name` and `ns` pointers). I am fine with changing that to be value based to make them stable across invocations (which is part of what this PR does). However, I think we should keep hashes for mutable Go object wrappers (including readers and writers) based on pointers. I agree it's safer to pre-calculate such hashes during object construction to avoid issues with GC moving objects around. Better yet, hash calculation can be done lazily (as you pointed out), but that part doesn't have to be part of this PR.

jcburley:
That all makes sense to me, and simplifies the PR/code even further -- pushed!

candid82:
Thanks a lot!

PR #313, opened 2020-01-04
Clean up and clarify various library issues

jcburley:
`./joker -e "(use 'joker.repl)"` wasn't working, as it turns out that my earlier "fix" to added that namespace to `*loaded-libs*` was wrong -- that should contain only std-generated libraries, not core libraries.

Fixed that, removed `joker.core` from that list as well (not that it should matter), and clarified the docs.

While at it, noticed the latest library was missing `:added` tags, so added them.

Sorted the list of core libraries for which docs are generated, and made that snippet into a test, to catch regressions.


candid82:
Sorry for the delay, I was on vacation. Thanks a lot!

PR #317, opened 2020-01-12
Add *core-namespaces* variable and improve *loaded-libs*

jcburley:
Having `*core-namespaces*` makes enumerating the core libraries straightforward -- no more having to list them separately when generating docs or testing that they can be loaded.

Also, ensure `joker.core` gets added to `*loaded-libs*` when it is "processed" (as it always is).

Finally, clarify what `*loaded-libs*` represents.


candid82:
Thanks!

PR #318, opened 2020-01-13
Add joker.core/go-spew private function

jcburley:
This has proven to be invaluable in debugging certain aspects of Joker internals. It's not intended for use by Joker users.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -13,6 +13,9 @@ jobs:
           name: go get profile
           command: go get -d -v github.com/pkg/profile
       - run:
+          name: go get go-spew
+          command: go get -d -v github.com/jcburley/go-spew
+      - run:
           name: go get yaml
           command: go get -d -v gopkg.in/yaml.v2
       - run:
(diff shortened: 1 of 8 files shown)


candid82:
While I can appreciate the convenience of having such function, I am not 100% convinced it's valuable enough to justify additional dependency or a lot of additional code.  Do you think it could be conditionally compiled only with specific build tag (similar to what's done for plan9)? That way folks who need it could still get it, but default build would not include this code or the dependency...

jcburley:
> While I can appreciate the convenience of having such function, I am not 100% convinced it's valuable enough to justify additional dependency or a lot of additional code. Do you think it could be conditionally compiled only with specific build tag (similar to what's done for plan9)? That way folks who need it could still get it, but default build would not include this code or the dependency...

Cool idea, I'll look into doing just that!

jcburley:
Ready to go!

jcburley:
Really truly ready to go!

Although, arguably the config map (or nil) should be the first argument, and thing(s) to be dumped would come after that...if this was a public function, I'd be more concerned about getting something like this right the first time. But let me know if you want that change made.

candid82:
I think it's good as is. Thanks a lot!

Issue #315, opened 2020-01-04
Eval test `deps-test.joke` fails in Git Bash on Windows


jcburley:
This seems to be the result of a discrepancy (or more than one?) regarding whether to use forward versus backwards slashes, something I've seen in Go commands before in this environment:

```
Running test in subdirectory classpath
Running test in subdirectory large-forked-stdout
Running test in subdirectory load-file
Running test in subdirectory multi-methods
Running test in subdirectory read-line
Running test in subdirectory stdin-pipe
<joker.core>:3534:3: Eval error: <joker.core>:3551:13: Eval error: open lib\test-local\lib.joke: The system cannot find the path specified.
Stacktrace:
  global tests/run-eval-tests.joke:118:1
  joker.test.run-eval-tests/main tests/run-eval-tests.joke:109:29
  joker.test.run-eval-tests/run-internal-tests tests/run-eval-tests.joke:88:4
  joker.test.run-eval-tests/run-internal-tests tests/run-eval-tests.joke:98:16
  core/apply <joker.core>:525:4
  core/apply__ tests/run-eval-tests.joke:95:23
  core/with-bindings* <joker.core>:1408:7
  core/apply <joker.core>:525:4
  core/apply__ tests/run-eval-tests.joke:96:25
  core/load-file tests/eval/deps-test.joke:1:1
  core/require <joker.core>:3500:3
  core/apply <joker.core>:527:4
  core/apply__ <joker.core>:3417:9
  core/apply <joker.core>:527:4
  core/apply__ <joker.core>:3370:5
  core/with-bindings* <joker.core>:1408:7
  core/apply <joker.core>:525:4
  core/apply__ <joker.core>:3373:11
  load <joker.core>:3313:3
  core/load <joker.core>:3547:9
  core/with-bindings* <joker.core>:1408:7
  core/apply <joker.core>:525:4
  core/apply__ <joker.core>:3551:13
Stacktrace:
  global tests/run-eval-tests.joke:118:1
  joker.test.run-eval-tests/main tests/run-eval-tests.joke:109:29
  joker.test.run-eval-tests/run-internal-tests tests/run-eval-tests.joke:88:4
  joker.test.run-eval-tests/run-internal-tests tests/run-eval-tests.joke:98:16
  core/apply <joker.core>:525:4
  core/apply__ tests/run-eval-tests.joke:95:23
  core/with-bindings* <joker.core>:1408:7
  core/apply <joker.core>:525:4
[...]

PR #319, opened 2020-01-13
Add re-quoted and re-string to convert between String and Regex

jcburley:
This will enable a more-elegant fix for `deps-test.joke` under Windows (Git Bash).

Driven by https://github.com/candid82/joker/issues/315, but not a fix for that, which will follow once this is merged.


candid82:
I am hesitant to add more arbitrary stuff to core namespace. Can those functions be moved to joker.string instead?

jcburley:
> I am hesitant to add more arbitrary stuff to core namespace. Can those functions be moved to joker.string instead?

Sounds good to me. I went with `joker.core` only because of the existing `re-*` functions there.

(I wonder if maybe we should create a `joker.regexp` namespace for these two new ones? And maybe move the existing core ones over to it? Just spitballing here...they're separate in the Go library, after all....)

jcburley:
Though I remain a tad concerned about `re-*` functions being split between `joker.core` and `joker.string`, it does make sense for the current set to be in core since that's how Clojure does it.

So I went ahead and moved the two new ones into `joker.string` as you suggested.

Adding tests, I discovered that comparing `Regex` objects doesn't really work if they're distinct objects but otherwise identical. I think this latest change fixes that, by comparing the original strings used to compile them. (Go `regexp` doesn't define an Equality function nor receiver for `Regexp`.)

core/object.go
```diff
@@ -1356,7 +1360,7 @@ func (rx Regex) Print(w io.Writer, printReadably bool) {
 func (rx Regex) Equals(other interface{}) bool {
 	switch other := other.(type) {
 	case Regex:
-		return rx.R == other.R
+		return rx.R == other.R || rx.R.String() == other.R.String()
```
candid82: This is not how regex equality works in Clojure:
```
user=> (= #"asdf" #"asdf")
false
```
So I think the current version is correct.

jcburley: > This is not how regex equality works in Clojure:
> 
> ```
> user=> (= #"asdf" #"asdf")
> false
> ```
> 
> So I think the current version is correct.

Oops, didn't think to check that! Fixed.

And this led me to notice that `(joker.string/re-string r)` is just another way of saying `(str r)`, i.e. Joker users already have that function.

So that leaves only `joker.string/re-quoted` as new functionality here (plus the tests).

Is it still worth introducing `joker.string/re-string`? If not, should `re-quoted` be moved back into `joker.core` for consistency with the others?

jcburley:
Hmm, Clojure already does offer this, which is different from `re-quoted`, which it doesn't not appear to offer:

https://clojuredocs.org/clojure.string/re-quote-replacement

A few observations:

* It's in `clojure.string`, not `clojure.core`, which I think answers the question (which had just occurred to me) as to where `re-quoted` would go if Clojure was to add it
* Maybe Joker should offer `re-quote-replacement` as well?
* If Joker ever does offer that, or Clojure offers `re-quoted` functionality, a better name for it might be `re-quote-string` or `re-quote-match`, to be consistent with `re-quote-replacement`?

jcburley:
It looks like Clojure does offer something akin to `re-quoted` in a roundabout way:

```
user=> (java.util.regex.Pattern/quote "a.b")
"\\Qa.b\\E"
user=>
```

It doesn't support ctor-style construction though:

```
user=> (java.util.regex.Pattern. "abc.def")
CompilerException java.lang.IllegalArgumentException: No matching ctor found for class java.util.regex.Pattern, compiling:(NO_SOURCE_PATH:13:1) 
user=>
```

I'm beginning to think maybe this PR should be shelved until we figure out the "right" way to offer `re-quoted` functionality (since `re-string` is already there as `str`).

In the meantime, for Windows' sake, `joker.string/replace` and such could be enhanced to support `String` as well as `Regex` types.

candid82:
I propose to remove `re-string` as `str` does the same thing and rename `re-quoted` to `re-quote` to be consistent with both Clojure/Java and Go naming.

jcburley:
> I propose to remove `re-string` as `str` does the same thing and rename `re-quoted` to `re-quote` to be consistent with both Clojure/Java and Go naming.

Done!

candid82:
Thanks!

PR #322, opened 2020-01-15
Support joker.string/split string separator

jcburley:
This extends `split` beyond what `clojure.string/split` provides, makes `split` more consistent with other `joker.string` functions, and sets the stage for a straightforward fix to the Windows bug in `deps.joke`.

With this, https://github.com/candid82/joker/pull/319 is (for me) less urgent.


candid82:
This makes sense. Thanks!

PR #323, opened 2020-01-15
Document more types

jcburley:
In particularly, `(doc Proc)`, as a Slack user was asking how to use one of these.


candid82:
Thanks!

PR #324, opened 2020-01-17
Doc all interns

jcburley:
I found this handy; not sure anyone else would want it, but it was helpful to verify that the fast-startup version of Joker didn't generate any differences in docs versus the normal build of Joker (in my `gen-code` branch).


candid82:
Thanks!

PR #328, opened 2020-01-20
Clean up the doc script a bit

jcburley:
When I merged master into my gostd branch, my recent change (to master's generate-docs.joke) looked really ugly, with its conditional definition of a top-level function.

This cleans that up, leaving the code a tad more Clojure-like.


candid82:
Thanks!

PR #332, opened 2020-02-01
Lazily load core libs

jcburley:
This cherry-picks several commits from `gen-code` and follows up with subsequent hand-ported changes from that branch.

I'm not entirely sure this is the right approach, but thought it'd be a good start to this process to see how it goes and consider whether to continue trying to cherry-pick commits (where feasible) or go entirely with the hand-porting approach to which I'm accustomed.


jcburley:
On 2nd thought, since verbosity doesn't rely on this feature, I'll withdraw this PR and submit verbosity first. (Verbosity improvements have helped me greatly with this effort, but aren't strictly necessary; yet much of my new code has verbosity checks intertwined, as one might expect.)

PR #333, opened 2020-02-01
Bring verbosity improvements over from gen-code branch

jcburley:
This is not strictly necessary, but I found it invaluable, mainly when adding tests for verbosity levels greater than 1 (the default for one specification of `--verbose`) that would use `go-spew` to output substantial data sets.

### DIFF
--- a/core/data/core.joke
+++ b/core/data/core.joke
@@ -3222,11 +3222,19 @@
      (when-not (bound? v#)
        (def ~name ~expr))))
 
+(defonce ^:dynamic
+  ^{:private true
+    :doc "True while a verbose load is pending"}
+  *loading-verbosely* false)
+
 (defn in-ns
   "Sets *ns* to the namespace named by the symbol, creating it if needed."
   {:added "1.0"}
   ^Namespace [^Symbol name]
-  (var-set #'joker.core/*ns* (joker.core/create-ns name)))
+  (let [new-ns (joker.core/create-ns name)]
+    (when *loading-verbosely*
+      (printf "joker.core/in-ns: changing from %s to %s\n" joker.core/*ns* new-ns) (flush))
+    (var-set #'joker.core/*ns* new-ns)))
 
 (defonce ^:dynamic
   ^{:private true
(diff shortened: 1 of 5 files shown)


core/data/core.joke
```diff
@@ -3851,6 +3854,33 @@
...
+(add-doc-and-meta *verbose*
+  "Whether --verbose (or equivalent) was specified by the Joker invocation.
+
+  Note: changing this (via e.g. var-set) affects only running Joker code; it
+  does not change the value used by the Go code that implements Joker. If
+  needed, it's probably better to add a set-verbose function that changes
+  both, rather than changing Joker's internals so this variable is always checked."
+  {:added "1.0"})
+
+(add-doc-and-meta *verbosity-level*
```
candid82: Having both `*verbose*` and `*verbosity-level*` seems redundant to me as `*verbosity-level*` provides a superset of possible options compared to `*verbose*`. Also, I am not sure these need to be exposed as vars. I would rather just have private `(verbosity-level__)` function that just returns `VerbosityLevel` value.

jcburley: > Having both `*verbose*` and `*verbosity-level*` seems redundant to me as `*verbosity-level*` provides a superset of possible options compared to `*verbose*`. Also, I am not sure these need to be exposed as vars. I would rather just have private `(verbosity-level__)` function that just returns `VerbosityLevel` value.

Sounds good. I tend to default to more flexibility, but in this case it's not really warranted and actually (per the docstring) confusing. Will update PR as you suggest.

core/environment.go
```diff
@@ -28,6 +27,8 @@ type (
 		MainFile      *Var
 		args          *Var
 		classPath     *Var
+		verbose       *Var
```
candid82: This var is not used anymore and can be deleted, right?

jcburley:
Almost certainly! Not able to do this myself just now, but feel free to fix it yourself if you like, or I'll get to it later.

On February 2, 2020 4:19:31 PM EST, Roman Bataev <notifications@github.com> wrote:
>candid82 commented on this pull request.
>
>
>
>> @@ -28,6 +27,8 @@ type (
> 		MainFile      *Var
> 		args          *Var
> 		classPath     *Var
>+		verbose       *Var
>
>This var is not used anymore and can be deleted, right?
>
>-- 
>https://github.com/candid82/joker/pull/333#pullrequestreview-351968156

-- 

core/environment.go
jcburley: No, and neither is `verbosity` needed, nor the code that initializes them.

Gone!

candid82:
Thanks a lot!

PR #335, opened 2020-02-01
Exclude generated code from builds of code-generating tools

jcburley:
This avoids a chicken-and-egg problem that comes up when e.g. changing whether a Joker type is a reference type.

It seems better to just avoid the problem, by not including code being generated via `go generate ./...` when building the tools doing the generation, than to document how to work around it.


jcburley:
(Perhaps unsurprisingly, I've been bitten by this very problem often enough to want to solve it!)

jcburley:
The checks fail because `cmd/go` seems to require the list of tags to be space-separated; but Go version 13 requires them to be comma-separated.

jcburley:
Adding a commit to this PR so Go v1.13.7 (instead of 1.12.6) is used fixes the problem.

I'm not sure whether that's the _right_ fix, however....

jcburley:
If we want to continue support for Go v1.12.x, these `go run` commands could be put in a shell script and those run instead, with the script choosing to comma-separate or space-separate based on the Go version.

Not sure that's worth doing versus just requiring Go v1.13.x going forward....

Updated the `README.md` in case this approach is approved.

jcburley:
Not sure why I thought this was an issue, but now I reailze `gen_types.go` does _not_ import `core` at all, so it can't run into this problem in the first place...!

jcburley:
And though `gen_data.go` does import `core`, the other PR I submitted earlier today, which deletes all the `core/a_*.go` files before running `go generate ./...`, should avoid problems in that area.

PR #337, opened 2020-02-03
Make Proc's self-identify

jcburley:
Needed by the `gen-code` branch (on my fork) to generate static Go code that instantiates a `Proc` given one that it finds while traversing Joker's data structures.

Also makes this work:

```
$ ./joker
Welcome to joker v0.14.1. Use EOF (Ctrl-D) or SIGINT (Ctrl-C) to exit.
user=> cons
#object[Proc:procCons]
user=> joker.string/blank?
#object[Proc:std/string.isblank_]
user=>
```

Previously, only `#object[Proc]` would be output, providing no information on which `Proc` was stringized.

(Note that the individual commit messages are missing the `#object` lines, presumably because they were treated as comment lines!)

### DIFF
--- a/core/object.go
+++ b/core/object.go
@@ -402,7 +402,6 @@ func init() {
 		NodeSeq:       RegRefType("NodeSeq", (*NodeSeq)(nil), ""),
 		ParseError:    RegRefType("ParseError", (*ParseError)(nil), ""),
 		Proc:          RegRefType("Proc", (*Proc)(nil), "A self-identifying callable function implemented via Go code"),
-		ProcFn:        RegRefType("ProcFn", (*ProcFn)(nil), "A callable function implemented via Go code"),
 		Ratio:         RegRefType("Ratio", (*Ratio)(nil), "Wraps the Go 'math.big/Rat' type"),
 		RecurBindings: RegRefType("RecurBindings", (*RecurBindings)(nil), ""),
 		Regex:         RegRefType("Regex", (*Regex)(nil), "Wraps the Go 'regexp.Regexp' type"),
--- a/core/object.go
+++ b/core/object.go
@@ -149,8 +149,9 @@ type (
 	}
 	ProcFn func([]Object) Object
 	Proc   struct {
-		Fn   ProcFn
-		Name string
+		Fn      ProcFn
+		Name    string
+		Package string // "' for core (this package), else e.g. "std/string"
 	}
 	Fn struct {
 		InfoHolder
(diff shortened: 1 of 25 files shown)


core/object.go
```diff
@@ -395,7 +401,8 @@ func init() {
 		Nil:           regType("Nil", (*Nil)(nil), "The 'nil' value"),
 		NodeSeq:       RegRefType("NodeSeq", (*NodeSeq)(nil), ""),
 		ParseError:    RegRefType("ParseError", (*ParseError)(nil), ""),
-		Proc:          RegRefType("Proc", (*Proc)(nil), "A callable function implemented via Go code"),
+		Proc:          RegRefType("Proc", (*Proc)(nil), "A self-identifying callable function implemented via Go code"),
+		ProcFn:        RegRefType("ProcFn", (*ProcFn)(nil), "A callable function implemented via Go code"),
```
candid82: Does `ProcFn` really need to be exposed as Joker type?

jcburley: > Does `ProcFn` really need to be exposed as Joker type?

I'm not sure. Originally it was needed because only the `core` procedures were converted over; I converted the `std` procedures over much later during this project.

The only unwrapped `ProcFn` of which I'm unaware is the lazy-seq-supporting one in `csv_native.go`. Offhand, I'm not sure whether that'd ever be exposed; I haven't investigated Joker's implementations of lazy sequences.

I'm okay with removing it; let me know if you want me to do that.

candid82: Yes, let's remove it. We can always add it back if needed.

jcburley: Done!

core/object.go
```diff
@@ -875,6 +885,42 @@ func (p Proc) GetType() *Type {
...
 
 func (p Proc) Hash() uint32 {
+	return p.Fn.Hash()
+}
+
+func (p ProcFn) Call(args []Object) Object {
+	return p(args)
+}
+
+func (p ProcFn) Compare(a, b Object) int {
```
candid82: One more thing: since we are not exposing ProcFn as Joker type, I don't think it needs to implement Object interface...

jcburley: I've removed all but the two receivers to which `Proc` delegates its implementations: `Equals()` and `Hash()`.

Want me to pull those implementations into `Proc`'s and remove them as well?

candid82: Yes, that would make more sense to me unless you have specific reasons to keep them in ProcFn.

jcburley: That's done, leaving only `(ProcFn) Call()`, which only `csv_native.go` uses.

Might be worth having that code wrap the `ProcFn` it builds in a `Proc` so the `(ProcFn) Call()` receiver can be eliminated as well -- will try that next.

jcburley: That builds and passes tests, but offhand I can't find any tests specific to `csv.joke`, so am not sure I haven't broken something here. Looking into to coming up with a decent test....

jcburley: Okay, that seems to work just fine.

candid82:
Thanks a lot!

PR #338, opened 2020-02-03
Separate environment.go init code from decl code

jcburley:
This is a precursor to separating out the initialization code into a distinct source file, `environment_init.go`, that (like the other `*_init.go` files that will be split off) is built only when `!fast_init` (which is true when building the normal version of Joker).

The fast-startup version of Joker is built with `-tags fast_init`, so those `*_init.go` files will not be compiled in, but their effects will be in the form of `core/a_*code.go` files generated by `gen_code.go` (forthcoming).

(It seemed like a good idea to make this a separate PR from the PR to actually split out the files, which will make it hard to detect changes in the text of the code that is split out. The idea here is to have no such changes to that code, just the source-file splits, in a subsequent PR.)

### DIFF
--- a/core/environment.go
+++ b/core/environment.go
@@ -59,6 +59,8 @@ func (env *Env) SetEnvArgs(newArgs []string) {
 	}
 }
 
+/* This runs after invariant initialization, which includes calling
+   NewEnv().  */
 func (env *Env) SetClassPath(cp string) {
 	cpArray := filepath.SplitList(cp)
 	cpVec := EmptyVector()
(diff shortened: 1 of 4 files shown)


core/environment.go
```diff
@@ -244,3 +265,7 @@ func (env *Env) ResolveSymbol(s Symbol) Symbol {
 		ns:   vr.ns.Name.name,
 	}
 }
+
+func init() {
+	GLOBAL_ENV.SetCurrentNamespace(GLOBAL_ENV.EnsureNamespace(MakeSymbol("user")))
```
candid82: I am not seeing a corresponding removal of a similar line anywhere in the diff, so not sure what it's for...

jcburley: > I am not seeing a corresponding removal of a similar line anywhere in the diff, so not sure what it's for...

`NewEnv()` used to do this (to `currentNs`), though it's not immediately obvious -- see the combination of the `NewEnv()` call in the `var GLOBAL_ENV` definition and line 88 (I think it is) in the old `NewEnv()`.

Without this line (all of `func init()` here, that is), the build quickly crashes.

candid82: Gotcha!

candid82:
Thanks!

PR #339, opened 2020-02-04
Split std-generated a_*.go files' init code into a_*_slow_init.go files

jcburley:
Also prep for ultimately merging the `gen-code` branch, this separates out static from dynamic initialization code so that std libraries, needed by core Joker namespaces, can participate in the fast-startup-initialization mechanism.

For examples, `joker.string` is required by `joker.repl`, and `joker.html` is required by `joker.hiccup`.

When those core libraries (`joker.repl` and `joker.hiccup`) are compiled to static-initialization forms (for fast startup), those forms need to be able to refer to static-initialization forms for the std libraries upon which they depend (`joker.string` and `joker.html`).

Rather than trying to do this by hand, this commit adds dynamic discovery of such preloaded namespaces (`joker.string` and `joker.html`) as well as of all the std namespaces (versus a static list).

While this commit doesn't employ the resulting set of preloaded namespaces (no fast-startup-only files, which will be named `a_*_fast_init.go`, are as yet generated), it does generate separate `a_*_slow_init.go` files. These perform runtime initialization that will otherwise be done by the fast-startup files.

The new (private) function `joker.core/ns-initialized?` takes a Symbol, denoting a namespace by name, and returns whether it has been initialized. `generate-std.joke` uses that to find out which namespaces need to be preloaded (which will be important later on).


candid82:
Thanks!

PR #340, opened 2020-02-04
Move listing of core .joke files into procs.go from gen_data.go

jcburley:
This will allow `gen_code.go` to share the same list.


jcburley:
Whoops, based changes on wrong branch.

PR #341, opened 2020-02-04
Move listing of core .joke files into new gen_common pkg from gen_data.go

jcburley:
This will allow `gen_code.go` to share the same list.


candid82:
`CoreSourceFiles` seems out of place in `procs.go`. It doesn't need to be part of Joker executable, does it (since it's only needed for code generation)? So perhaps it should be move into its own `gen_common` package?

jcburley:
> `CoreSourceFiles` seems out of place in `procs.go`. It doesn't need to be part of Joker executable, does it (since it's only needed for code generation)? So perhaps it should be move into its own `gen_common` package?

Great idea!! Done.

candid82:
Thanks!

PR #342, opened 2020-02-05
Split core files

jcburley:
This should just split the runtime-initialization code (in `core`) out into `*_init.go` files, without making any changes to the code.

Those new files are built only when not `-tags fast_init`; when in that mode, their effects are implemented via static code generated by `gen_code` and compiled into the Joker executable.


candid82:
Thanks!

PR #343, opened 2020-02-05
Handy reference right in the script

jcburley:
Alternately, this could be reformatted and put in `DEVELOPER.md` with a suitable URL at the top of this script?


candid82:
Yeah, I think I like the idea of moving this to `DEVELOPER.md`. Otherwise the code becomes too cluttered with non-code.

jcburley:
> Yeah, I think I like the idea of moving this to `DEVELOPER.md`. Otherwise the code becomes too cluttered with non-code.

Makes sense -- done!

candid82:
Thanks!

Issue #344, opened 2020-02-05
Dynamically Discover Core Libraries?


jcburley:
Rather than maintain static lists of core namespaces and their source files in `core/data/`, consider structuring that directory as per usual Joker namespace rules and having `gen_data` (and/or `gen_code`) dynamically discover and process them.

`core/data/joker/core.joke` would still be loaded first and treated specially, but the rest would be loaded as needed based on the usual `(ns ...)` specifications.

While this would make official Joker more maintainable for the occasional addition of a core namespace, it could make Joker much more attractive to end users who wish to "compile", or embed, their own (or 3rd-party) libraries in the executable -- they could just drop the source code into the appropriate directory and rebuild.

harold:
An interesting idea!

Issue #326, opened 2020-01-18
Map All Builtin (Core) Namespaces?


jcburley:
It might simplify things to automatically map all the namespaces compiled (built) into Joker from their source files in `core/data/*.joke`.

They start out mapped in the `-tags fast_init` build of the `gen-code` branch on my fork, and that doesn't seem to hurt anything.

If `gen_data.go` (say) emitted the necessary calls to add generated namespaces to `GLOBAL_ENV.Namespaces`, it might then be possible to not only simplify the documentation, but simplify or eliminate the definition of `joker.core/*loaded-libs*`, the `internalLibs` var in `package core` and such.

jcburley:
This is now done by the PR(s) referenced above.

PR #345, opened 2020-02-05
Split out []byte-array generation for core-library namespaces

jcburley:
This eliminates the need to hand-maintain the declaration of the `[]byte` arrays, allowing `gen_data.go` to automatically generate those declarations and initialization (statically, in a func init(), or in a lazily-called function) as it sees fit. `internalLibs` would be empty for `gen_data`, but that seems appropriate, and it works fine.

Something like this might be a next step towards addressing https://github.com/candid82/joker/issues/326 and/or https://github.com/candid82/joker/issues/344.

(`data.go` might be a better name than `procs_data.go`?)


candid82:
> (data.go might be a better name than procs_data.go?)

Not sure... I am not even sure it's worth creating a separate file for those functions...Granted `procs.go` has become a kitchen sink but I am not sure if many small files are necessarily better. What do you think?

candid82:
I guess `procs_data.go` would be excluded in fast_init build? In that case it does make sense to have it as a separate file...

jcburley:
> I guess `procs_data.go` would be excluded in fast_init build? In that case it does make sense to have it as a separate file...

Actually, it's excluded when building/running `gen_data` -- that's what drove moving that stuff there. Else there'd be conflicts.

I think I prefer the name `data.go` for it though.

candid82:
I saw that it's excluded in `gen_data` build, but I did not realize it was *necessary* to exclude it. What conflicts would it cause?

jcburley:
> I saw that it's excluded in `gen_data` build, but I did not realize it was _necessary_ to exclude it. What conflicts would it cause?

"Conflicts" was the wrong word, actually.

Since `gen_data` generates the _declarations_ of `[]byte` vars, the code in `data.go` won't build until `gen_data` has run.

This might seem a bit strange for now, but I've got another PR following it up that removes all but the references to the linter `[]byte` vars from `data.go`.

candid82:
Ah, that makes sense. Thanks!

PR #347, opened 2020-02-06
Map core namespaces

jcburley:
This implements https://github.com/candid82/joker/issues/326, making things a bit simpler and more consistent overall (less manual work when adding a new library, fewer visible differences among core, std, and external libs, etc).


candid82:
Thanks!

PR #350, opened 2020-02-15
Create certain std/*/*_fast_init.go files

jcburley:
Though not yet needed, the `gen-code` branch uses these to build the fast-startup version of Joker.

Only the std libs (namespaces) depended upon by the core Joker libraries get this treatment (i.e. get `a_*_fast_init.go` files); the others are unaffected.


candid82:
Thanks!

PR #351, opened 2020-02-15
Refer core namespace to user namespace only when building slow version

jcburley:
This isn't needed for the fast version (forthcoming `gen-code` branch), nor by `gen_data.go` that I can see.

### DIFF
--- a/main.go
+++ b/main.go
@@ -690,8 +690,7 @@ func main() {
 	RT.GIL.Lock()
 	ProcessCoreData()
 
-	GLOBAL_ENV.FindNamespace(MakeSymbol("user")).ReferAll(GLOBAL_ENV.CoreNamespace)
-
+	GLOBAL_ENV.ReferCoreToUser()
 	GLOBAL_ENV.SetEnvArgs(remainingArgs)
 	GLOBAL_ENV.SetClassPath(classPath)
 
--- a/core/environment_init.go
+++ b/core/environment_init.go
@@ -40,3 +40,7 @@ func NewEnv() *Env {
 		MakeMeta(nil, "Map of configuration key/value pairs for linter mode", "1.0"))
 	return res
 }
+
+func (env *Env) ReferCoreToUser() {
+	env.FindNamespace(MakeSymbol("user")).ReferAll(env.CoreNamespace)
+}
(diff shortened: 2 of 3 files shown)


main.go
```diff
@@ -690,8 +690,7 @@ func main() {
 	RT.GIL.Lock()
 	ProcessCoreData()
 
-	GLOBAL_ENV.FindNamespace(MakeSymbol("user")).ReferAll(GLOBAL_ENV.CoreNamespace)
-
+	GLOBAL_ENV.ReferCoreToUser()
```
candid82: Won't this fail compilation with `fast_init` tag since `ReferCoreToUser` is not available with `fast_init`?

jcburley: `gen_code.go` (which generates the `fast_init` stuff, so that doesn't actually work yet on `master` anyway) generates an empty `ReferCoreToUser` implementation into `a_code.go`. So this does work on the `gen-code` branch.

That's arguably a bit of a kludge, in that it _could_ go into a proper `*.go` file compiled only with `+build fast_init`.

But right now there's not really any such "source" file in the ordinary sense, so I threw it into the generated `a_code.go` file.

Do you want it somewhere else? E.g. a `core/fast_init.go` file, with the "slow" version going into `core/slow_init.go`? They'd start out with just this one receiver in them, but might grow over time to include more stuff that isn't really specific to other existing files.

Or, we could put it in `core/environment_fast_init.go` and rename `core/environment_init.go` to `core/environment_slow_init.go` (and maybe the other `core/*_init.go` files accordingly, for consistency).

(As an aside, one of the things I absolutely adore about Go is the ability to do such simple renaming without having to edit Makefiles....)

jcburley: Upon further consideration, I find I'm really liking the idea of all `*_init.go` files being either `*_slow_init.go` or `*_fast_init.go`.

I've updated my pending `gen-code` branch to put the empty implementation of this new receiver into `core/environment_fast_init.go`.

Then, unless I hear otherwise from you, I'll go forward with a (separate) renaming PR, that also adds the new source file from the `gen-code` branch.

But this current PR should be fine as-is, in the sense that it won't break anything.

candid82: This makes sense, thanks!

PR #352, opened 2020-02-15
Move the imports of std libraries to the new-ish gen_common.go file

jcburley:
As `gen_common.go` will be shared by `gen_data.go` and the forthcoming `gen_code.go`, the imports might as well be there so only one file need be changed when adding new std libraries needed by core libraries.

Update `DEVELOPER.md` accordingly and per previous PRs.


candid82:
Thanks!

PR #354, opened 2020-02-17
Add fast (empty) receiver and rename core/ files *_init.go to *_slow_…

jcburley:
…`*_slow_init.go`.

This leaves Joker with no `*_init.go` files that are not either `*_slow_init.go` or `*_fast_init.go` (and the latter are always tagged accordingly, while the former are when it's necessary to exclude them for the fast builds that the `gen-code` branch, a PR yet to come, supports).


candid82:
Oops, conflicts...

jcburley:
Will fix now.

jcburley:
Fixed!

candid82:
Thanks!

PR #355, opened 2020-02-17
Run go fmt on generated std/ files

jcburley:
Recent changes to how these files are generated do not consistently conform to Go coding conventions.

Rather than try to keep up with those via changes to `generate-std.joke` and its `*.tmpl` files, just run `go fmt ./...` on the `std/` directory after generating those files.

That way, mere formatting differences won't force rebuilding Joker.


candid82:
Thanks!

PR #358, opened 2020-02-19
Support and enable fast initialization to speed up Joker startup

jcburley:
Here is the "sanitized" `gen-code` branch, effectively squashed to a single commit.

In the past few weeks, I've refactored it to comprise two new packages (adding `gen_go`) consisting of several source files.

The result should be easier to understand, verify, and maintain.

Besides whatever issues are found in the code itself, I think one of the biggest questions is whether `run.sh` should (as it does here) default to not only generating the fast-initialization code (built with `-tags fast_init`), but also making the "fast" version of `joker` the default, as well as leaving `joker.slow` and `joker.fast` (a hardlink to `joker`) around.

The changes herein to `run.sh` might thus be deemed to aggressive, though given the monthly release schedule and the fact that it's now just a little over halfway through the month, giving bleeding-edge types a couple of weeks to vet this fast version, that might be okay.


candid82:
There is too much code to review line by line, but I understand the main idea and the end result looks great! I hope to deepen my understanding of this new code as I start working on enhancements / bug fixes. 

As for `run.sh` defaults, I think it's fine (and even desirable) to make the fast build a default one, as we want to give it as much testing "in the field" as possible before the next release. I think the end goal is to make it the *only* build pipeline and eliminate `gen_data.go` completely. But first we need to take it for a test drive :)

Thanks you very much, this is stellar work!

jcburley:
Wow, thanks! Will close the Issue (for now anyway -- maybe there are further improvements to be had, but I'm not aware of any that don't require changes to Go itself).

Issue #357, opened 2020-02-18
Crash if namespace in a file has too many segments?


raymond-w-ko:
[joker-crash.log](https://github.com/candid82/joker/files/4221064/joker-crash.log)

I am running the latest version of joker like `joker --lint tmp.cljs` in the directory `/tmp/tmp`.  If I remove the `.core` then it works. If I keep it as it then I get this long stack trace attached.

Is there some limit to namespace segment depth in Clojure and/or Java?

```clojure
(ns foo.bar.fizz.buzz.core
  (:require [clojure.string :as str]))

(+ 1 foo)
```

raymond-w-ko:
I don't know golang, but I suspect it has something to with `joker/core/procs.go:1495`

My use case is that I have a project where the namespaces are around 5 segments. I use vim and the vim-ale plugin. It calls joker like this:
`['/usr/bin/zsh', '-c', 'joker --working-dir ''/home/rko/src/redacted/redacted/src/redacted/redacted/views/redacted/core.cljs'' --lint ''/tmp/v3fzYms/294/core.cljs''']`

I suspect it is misusing the `--working-dir` argument.

Since the asynchronous linting requires a copy of the file, it copies it to `/tmp`. However, there are more namespace segments in the `(ns)` header than in the path holding the file, causing this issue?

PR #359, opened 2020-02-19
Fix panic when exhausting pathname components for namespace

jcburley:
Fixes crash on this when in a "short" path such as `/tmp`:

```
(ns foo.barfizz.buzz.core (:require [clojure.string :as str]))
```

Fixes: https://github.com/candid82/joker/issues/357


candid82:
Thanks a lot!

PR #361, opened 2020-02-22
Invert and rename build tag so -tags slow_init builds slow version

jcburley:
This changes `go build`, `go install`, etc. to default to building the fast version.


jcburley:
Not sure when is best to pull this in -- give some time for the current approach to bake first?

candid82:
I am not sure if this step is necessary. I was planning on keeping things as is for some time and then getting rid of slow_init altogether. I'll keep this PR open for now in case the plan changes for whatever reason. Thanks!

jcburley:
> I am not sure if this step is necessary. I was planning on keeping things as is for some time and then getting rid of slow_init altogether. I'll keep this PR open for now in case the plan changes for whatever reason. Thanks!

That's fine. Just wanted to give you the option of ensuring that "default" builds (e.g. situations where people didn't notice they should do `go build -tags fast_init` or `go install -tags fast_init` by hand or in their build processes) picked up, and thus tested, the "fast" version, before we removed the ability to easily do A/B testing on a single version of Joker.

jcburley:
Note that I started working on removing the `fast_init` stuff first, so only a single (fast-startup) executable would get built.

That's more involved, of course, than this PR.

At first I thought it'd be straightforward to get rid of `PackReader()` and such. Then I realized it'd still be needed for the `core/data/linter_*.joke` files, which are not handled by `gen_code`. It's not obvious to me what should be done with those.

jcburley:
> At first I thought it'd be straightforward to get rid of `PackReader()` and such. Then I realized it'd still be needed for the `core/data/linter_*.joke` files, which are not handled by `gen_code`. It's not obvious to me what should be done with those.

More discussion of this here: https://github.com/candid82/joker/issues/362

candid82:
I suppose there is no harm in merging this. Thanks!

Issue #362, opened 2020-02-24
Single Build Pipeline


jcburley:
Since merging the `fast_init` PR (https://github.com/candid82/joker/pull/358), Joker builds two versions of itself, switched on the presence/absence of the `fast_init` tag (or, with https://github.com/candid82/joker/pull/361, the `slow_init` tag).

When the "fast-startup" version is deemed sufficiently worthy, the "slow-startup" (original) version can and should be decommissioned.

This Issue is intended to capture what this entails -- and what it doesn't.

jcburley:
The `core/*_slow_init.go` files will still be needed (as will their build tag), because they contain the "original" source code for much of what `gen_code` is compiled with to, in turn, generate (compile) the `core/a_*code.go` files (compiled into Joker itself) that make the fast-startup stuff work.

Further, the `core/a_linter*_data.go` files (conceptually, any `core/a_*_data.go` file without a corresponding `core/a_*_code.go` file, and thus not conditionally built based on a `*_init` build tag, will also still be needed, unless some other mechanism is developed. That's because they have historically been read in only _conditionally_ at run time _and_ change core namespaces already established via "slow" or "fast" initialization.

However, some `core/a_*_data.go` files can go away, if they have a `*_init` built tag causing them to be ignored when building the (fast/soon-to-be-only) version of Joker.

That means `gen_data` would no longer generate them. It would generate only the `core/a_linter*_data.go` files.

On the `std` side, I think all the `a_*_init.go` files (slow _and_ fast) can be merged back into `a_*.go` files. Where `a_*_fast_init.go` files are currently being generated (`joker.html` and `joker.string`, because they're depended upon by core libraries), the code in corresponding `a_*_slow_init.go` can pretty much be deleted (not generated by `generate-std.joke`; otherwise, that code would just be merged in.

Of course, `run.sh` would no longer build two versions of the executable.

There might be glitches in this plan, because I haven't actually coded it up yet. It's probably best to wait for clarity on the viability of `joker.fast` (the fast-startup version) and a final decision as to whether https://github.com/candid82/joker/pull/358 will be merged, before going back to a single-build pipeline as per this Issue, to avoid messy merges down the road.

jcburley:
Once this is done, https://github.com/candid82/joker/issues/363 should be considered.

jcburley:
At this point, https://github.com/candid82/joker/pull/361 is probably the best next step, as it'd enable easily making small, incremental PRs towards resolving this Issue.

[...]

PR #364, opened 2020-02-26
Make gen-code Independent of gen-data

jcburley:
This seems like the logical next step towards resolving https://github.com/candid82/joker/issues/362.


candid82:
Thanks!

PR #365, opened 2020-02-27
Build only one executable via run.sh

jcburley:
The next step towards https://github.com/candid82/joker/issues/362, this causes `run.sh` to (in the default case) no longer build a `joker.slow` executable. Or, an env var or flag file can be used to build that but not `joker.fast`.

`go build -tags slow_init` and such can still be used afterwards to build such an executable (as `joker`), so A/B testing is still supported. That is, all the relevant `*_data.go`, `*_code.go`, `*_slow_init.go`, and `*_fast_init.go` files are still generated.

Documentation updated accordingly, including fixing/clarifying a few things that predate this PR.

### DIFF
--- a/DEVELOPER.md
+++ b/DEVELOPER.md
@@ -421,13 +421,13 @@ So, while `joker.repl` and `joker.tools.cli` currently depend on `joker.string`,
 
 ## Faster Startup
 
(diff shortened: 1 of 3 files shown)


candid82:
Thanks!

PR #366, opened 2020-02-28
Get joker.bolt working

jcburley:
Needs a test of some sort, which would have caught this.

(Also, would be great to automatically populate `*loaded-libs*` as is being done for `*core-namespaces*`.)


candid82:
Thanks! I am so used to the fact that built-in namespaces are available without requiring them that I forgot to actually test `(require 'joker.bolt)`.

PR #369, opened 2020-02-28
Automate populating *loaded-libs*

jcburley:
One more thing to not have to remember to do when adding a lib to `std/`....


candid82:
👍

PR #370, opened 2020-02-28
Add :ns and :name to metadata via add-doc-and-meta

jcburley:
Note the missing info in current Joker versus with this change:

```
$ joker -e '(joker.repl/doc *file*)'
-------------------------

  The path of the file being evaluated, as a String.

  When there is no file, e.g. in the REPL, the value is not defined.
$ ./joker -e '(joker.repl/doc *file*)'
-------------------------
joker.core/*file*
  The path of the file being evaluated, as a String.

  When there is no file, e.g. in the REPL, the value is not defined.
```


candid82:
Thanks!

PR #371, opened 2020-03-01
Add test for bolt.joke

jcburley:


jcburley:
This test pulls in several other dependencies (Joker namespaces), though it's not the only such test that pulls in at least one other than `joker.test`.

Not sure how problematic that is; I think having the test is better than not having it.

Though this might not be a great first case, someday we might want to have a heavier-weight test suite that isn't used to gate PRs.

candid82:
Thanks!

PR #378, opened 2020-03-08
Remove slow_init build tag

jcburley:
This is the next step towards resolving https://github.com/candid82/joker/issues/362.

It should be merged only after confidence in **v.0.15.0** is sufficiently high that the ability to do easy A/B testing of the slow and fast versions of Joker can be comfortably dropped, as this PR removes that ability.

Besides removing support for **slow_init**, some other build-tag stuff is cleaned up, and the ability to easily see what `generate-std.joke` is doing (a simple `(debug ...)` facility) is added.

The remaining work, after this is merged, is pretty much ready to go as a subsequent PR. They could be submitted together as a single PR, instead of this one, if desired -- but this seems large enough as it is.


candid82:
Thanks!

PR #379, opened 2020-03-09
Better bolt test fix

jcburley:
Close file ASAP (and nil out the var for it) while avoiding race conditions.

Before closing the file, capture its name, then use that when opening the db and, later, removing the file.

Tested on Windows (and Ubuntu).

### DIFF
--- a/tests/eval/bolt.joke
+++ b/tests/eval/bolt.joke
@@ -9,8 +9,8 @@
   (testing "example in docstring"
     (let [f (create-temp "" "bolt-test-")
           db-name (name f)
-          db (open db-name 0600)
           _ (joker.os/close f) ; Windows requires the file itself to be closed for remove to work.
+          db (open db-name 0600)
           users "users"
           joe "Joe Black"]
       (try
--- a/tests/eval/bolt.joke
+++ b/tests/eval/bolt.joke
@@ -8,8 +8,9 @@
 (deftest example-1
   (testing "example in docstring"
     (let [f (create-temp "" "bolt-test-")
-          db (open (name f) 0600)
-          _ (joker.os/close f)  ; Windows requires the file itself to be closed for remove to work.
+          db-name (name f)
+          db (open db-name 0600)
+          f (joker.os/close f) ; Windows requires the file itself to be closed for remove to work.
           users "users"
           joe "Joe Black"]
       (try
@@ -21,4 +22,4 @@
               s (str (sort m))]
           (is (= s (format "([\"id\" %d] [\"name\" \"%s\"])" id joe))))
         (finally (close db)
-                 (remove (name f)))))))
+                 (remove db-name))))))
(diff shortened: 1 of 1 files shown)


tests/eval/bolt.joke
```diff
@@ -8,7 +8,9 @@
 (deftest example-1
   (testing "example in docstring"
     (let [f (create-temp "" "bolt-test-")
-          db (open (name f) 0600)
+          db-name (name f)
+          db (open db-name 0600)
+          f (joker.os/close f) ; Windows requires the file itself to be closed for remove to work.
```
candid82: It would seem more logical to me if these lines were swapped, so that `open` doesn't try to open a file that we already hold an open descriptor for... Also `f` doesn't seem to be used after closing, so it can be `_ (joker.os/close f)` instead.

jcburley: Are you sure there can't be a collision (in theory anyway) if the file is closed before being reopened? I don't know (offhand) if the temp-name mechanism relies on O_EXCL or similar to ensure no collision. But closing the file first would be my preference as well, though in practice doing it after opening the db on it seems to work fine.

Also, I started with `_`, then realized it'd be cleaner to re-bind `f` to `nil` after closing the file on it. That's how I discovered the "dangling" `(name f)` evaluations. So I think it's better to stick with this as a sort of idiom (in the sense of this test possibly being used as sample code), even though it isn't needed here.

Please let me know what you'd like me to do here, or feel free to fix as you see fit.

candid82: It does [use](https://golang.org/src/io/ioutil/tempfile.go?s=1419:1477#L40) O_EXCL, so no collisions there.

What's the advantage of binding `f` to `nil` after closing the file? `(name f)` is legal (and works) after closing. If you try to read/write using `f` after closing you'll get `file already closed` error. Besides, technically the fact that `joker.os/close` returns `nil` is not documented and should not be relied upon.

jcburley: > It does [use](https://golang.org/src/io/ioutil/tempfile.go?s=1419:1477#L40) O_EXCL, so no collisions there.

Doesn't that mean closing the file essentially "releases" it for any other process to open for its own (temporary) use? I realize the window of time (before the db is opened on the file) is very small and this is very unlikely....

> What's the advantage of binding `f` to `nil` after closing the file? `(name f)` is legal (and works) after closing. If you try to read/write using `f` after closing you'll get `file already closed` error. Besides, technically the fact that `joker.os/close` returns `nil` is not documented and should not be relied upon.

Didn't realize `(name f)` worked on a closed file -- though I think the code makes sense as it currently stands _vis-a-vis_ capturing the name once and reusing that.

So, since `joker.os/close` returning `nil` is undefined, I'll change binding its result back to `_`.

jcburley: > > It does [use](https://golang.org/src/io/ioutil/tempfile.go?s=1419:1477#L40) O_EXCL, so no collisions there.
> 
> Doesn't that mean closing the file essentially "releases" it for any other process to open for its own (temporary) use? I realize the window of time (before the db is opened on the file) is very small and this is very unlikely....

Wait, I looked at the code, and it (of course) uses `O_CREAT` as well.

So this is a non-issue. Will change the test accordingly.

candid82:
Thanks!

PR #380, opened 2020-03-10
Document constants and improve doc contents

jcburley:
Tag and documents constants as such; add Contents sections to namespace docs; distinguish kinds (versus types) of things more clearly via colors and such.


jcburley:
Just a reminder: as this PR changes `main.css`, a "mere" reload/refresh of a doc page might not be enough to pick up changes to that. E.g. on Mac OS X, I use Shift-Command-R (instead of just Command-R) to get the latest `main.css`.

jcburley:
While working on further improvements, I noticed this was incorrectly rendered:

https://candid82.github.io/joker/joker.core.html#%3C!

Hence the latest commit (ae80e21).

Let me know if you want that as a stand-alone PR (which would conflict with this one).

jcburley:
I've merged this and a pending doc-gen change (to add type/tag info to `{usage}` substitutions) into my `gostd` work, generated the docs there, and uploaded as usual.

So you can get a quick idea of what this PR would change by looking at e.g.:

https://burleyarch.com/joker/docs/amd64-linux/joker.core.html

candid82:
Thank you for providing the link to an example, this is very helpful.

I think it's helpful to have separate sections for variables and constants. Perhaps it also
makes sense to move macros into a separate section as well. I do have a few reservations:
- Joker's (and Clojure's) special forms are not namespaced, so having 
"Functions, Macros, and Special Forms" section on namespace page doesn't make sense to me.
That is, special forms should not be part of a namespace page.
- Types are also not (yet) namespaced in Joker, so there should not be "Types" section on
namespace page. 
- Personal preference: I like the old minimalistic style of Index better 

I can merge and make the tweaks myself if you are OK with this.

jcburley:
> Thank you for providing the link to an example, this is very helpful.
> 
> I think it's helpful to have separate sections for variables and constants. Perhaps it also
> makes sense to move macros into a separate section as well.

I'm torn on that, because while it's certainly straightforward to do, I'm thinking that the typical reader isn't necessarily aware of whether something they're searching for (not by name, else they'd use the Index) is a macro or a function.

Whereas, the colorized highlighting of macros in the Index can help them quickly find just (say) Macros.

Either way is fine by me.

>I do have a few reservations:
> 
> * Joker's (and Clojure's) special forms are not namespaced, so having
>   "Functions, Macros, and Special Forms" section on namespace page doesn't make sense to me.

Currently they appear only for `joker.core`, and maybe they'd never appear anywhere else (not sure how they could!).

So I guess the main question is whether somebody expects to find e.g. `fn` documented in the `joker.core` namespace or will know to go to the top-level `index.html` to search for it.

I don't feel terribly strongly about this either.

> * Types are also not (yet) namespaced in Joker, so there should not be "Types" section on
>   namespace page.

Agreed. Not sure why I left that in there -- the `gostd` branch needs it, but not the official version.

> * Personal preference: I like the old minimalistic style of Index better

You mean without the boxes? I don't feel strongly about that. Just wanted to give it a try, I guess, and make items look more like the **Legend** items.

I do like the colorized highlighting (so it would be great if that change could be preserved), but it's also nice seeing so many more names on a given screen.

> I can merge and make the tweaks myself if you are OK with this.

Please feel free! Let me know if you want me to take care of any particular change here.

jcburley:
> > * Joker's (and Clojure's) special forms are not namespaced, so having
> >   "Functions, Macros, and Special Forms" section on namespace page doesn't make sense to me.
> 
> Currently they appear only for `joker.core`, and maybe they'd never appear anywhere else (not sure how they could!).
> 
> So I guess the main question is whether somebody expects to find e.g. `fn` documented in the `joker.core` namespace or will know to go to the top-level `index.html` to search for it.
> 
> I don't feel terribly strongly about this either.

Just checked -- they're present in the current official documentation. E.g.:

https://candid82.github.io/joker/joker.core.html#fn

Perhaps the phrase "and Special Forms" could be emitted only for `joker.core`? Or maybe special forms could just not be documented in the `joker.core` namespace itself?

candid82:
RE: special forms. It's a bit confusing because some "special forms" like `fn` are not really special forms (they are macros). I'll think about it more.

jcburley:
> > * Types are also not (yet) namespaced in Joker, so there should not be "Types" section on
> >   namespace page.
> 
> Agreed. Not sure why I left that in there -- the `gostd` branch needs it, but not the official version.

Ah, it's _not_ in the PR itself -- that webpage shows the `gostd` branch's documentation, which has numerous (auto-generated) types.

Sorry for the confusion!

PR #381, opened 2020-03-10
Finish folding gen_data, gen_common, etc into gen_code 

jcburley:
This should resolve https://github.com/candid82/joker/issues/362.


candid82:
Thanks!

PR #382, opened 2020-03-21
Exit REPL upon seeing :repl/quit

jcburley:
Scratching a longtime itch....

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ You can also [build](#building) Joker from the source code.
 
 ## Usage
 
-`joker` - launch REPL
+`joker` - launch REPL. Exit via `:repl/quit`, `EOF` (such as `Ctrl-D`), or `SIGINT` (such as `Ctrl-C`).
 
 `joker <filename>` - execute a script. Joker uses `.joke` filename extension. For example: `joker foo.joke`. Normally exits after executing the script, unless `--exit-to-repl` is specified before `--file <filename>`
 in which case drops into the REPL after the script is (successfully) executed. (Note use of `--file` in this case, to ensure `<filename>` is not treated as a `<socket>` specification for the repl.)
--- a/main.go
+++ b/main.go
@@ -154,6 +154,12 @@ func processReplCommand(reader *Reader, phase Phase, parseContext *ParseContext,
 		return
 	}
 
+	if kw, yes := obj.(Keyword); yes {
+		if kw.ToString(false) == ":repl/quit" {
+			ExitJoker(0)
+		}
+	}
+
 	if phase == READ {
 		fmt.Println(obj.ToString(true))
 		return false
@@ -220,7 +226,7 @@ func srepl(port string, phase Phase) {
 
 	reader := NewReader(runeReader, "<srepl>")
 
-	fmt.Fprintf(Stdout, "Welcome to joker %s, client at %s. Use '(joker.os/exit 0)', or close the connection, to exit.\n",
+	fmt.Fprintf(Stdout, "Welcome to joker %s, client at %s. Use ':repl/quit', or close the connection, to exit.\n",
 		VERSION, conn.RemoteAddr())
 
 	for {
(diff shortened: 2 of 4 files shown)


candid82:
This looks pretty arbitrary to me and it doesn't match `clojure/clj` behavior. What's the itch it's scratching?

jcburley:
> This looks pretty arbitrary to me and it doesn't match `clojure/clj` behavior. What's the itch it's scratching?

Clojure doesn't offer it (AFAIK), but lumo offers `:cljs/quit` (also `:cljs/exit` as a synonym) and babashka offers `:repl/quit`, which just seems like the most straightforward (you're in a REPL, so `:repl/` makes sense there).

Since I use Windows (where even `Ctrl-Z` doesn't work) and occasionally sRepl (where `Ctrl-D` and `Ctrl-C` don't work, at least not via telnet), and since `(joker.os/exit 0)` is kinda tedious, I thought having something might be nice, and have wanted it for months now.

candid82:
I see. What do you think about adding `(exit)` function to `repl.joke` instead? It seems more straightforward both conceptually (no magic keywords) and implementation wise.

jcburley:
> I see. What do you think about adding `(exit)` function to `repl.joke` instead? It seems more straightforward both conceptually (no magic keywords) and implementation wise.

If we go this route (add a new function), I think `joker.repl/quit` is a better name, so it doesn't conflict with `joker.os/exit` (as `(use joker.os)` is probably fairly popular in Joker's REPL).

I admit I'm not 100% clear why other Clojure implementations chose a special-cased keyword over a simple function invocation within their respective REPL (or `user`) namespace.

One thing I did notice, when implementing it, is that `:repl/quit` works even when `--read` is specified on the command line:

```
)$ ./joker --read
Welcome to joker v0.15.0. Use ':repl/quit', EOF (Ctrl-D), or SIGINT (Ctrl-C) to exit.
user=> (quit)
(quit)
user=> :repl/quit
$
```

But I don't know whether this (running the REPL in a mode other than EVAL) is even a use case worth worrying about, given Joker's maturity at this point. It might have been a factor while developing those other Clojure variants.

Also, though it'd be more special-casing, it'd be fairly easy to special-case `--read` and `--parse` so they recognized `(quit)`...if we ever had any reason to do that.

So `(quit)` would have been my original thought except for what other Clojure variants offer, and I'd be okay with that.

Let me know if you want me to implement that here (or in a new PR).

candid82:
I am fine with `(quit)`. I’ll implement it (probably later today) unless you beat me to it.

jcburley:
> I am fine with `(quit)`. I’ll implement it (probably later today) unless you beat me to it.

https://github.com/candid82/joker/pull/383

borkdude:
FWIW:

- `:cljs/quit` is supported by ClojureScript:

  ```
  $ clj -Sdeps '{:deps {org.clojure/clojurescript {:mvn/version "1.10.597"}}}' -m cljs.main -re node
  ClojureScript 1.10.597
  cljs.user=> :cljs/quit
  ```

- `:repl/quit` is also supported by [Rebel Readline](https://github.com/bhauman/rebel-readline).

- `lein repl` exits on the symbol `exit`

jcburley:
So keywords seem to be a bit more widespread.

But `lein repl` documents (on startup) `(quit)` and `(exit)`, while it also supports those as "commands" as well, as @borkdude points out regarding `exit` (so `quit` also works).

@candid82 -- based on what you're seeing/thinking, do you prefer keyword-based or function/command-based exiting? I don't have a preference, but think it'd be great for just `quit` or `exit` alone to work (i.e. as special-cased commands). Will comment accordingly on that other PR.

If you don't want keyword-based, feel free to close this PR.

candid82:
This is addressed by https://github.com/candid82/joker/commit/08a3b0c94cc5dda72a38e4c86d142cf4cdb5ef01 (I went with `joker.core/exit` approach)

candid82:
Thanks @jcburley for working through the different solutions and @borkdude for the input!

PR #383, opened 2020-03-23
Support (joker.repl/quit) to exit Joker

jcburley:
This does the job, but illustrates the awkwardness of defining a namespaced function for this, in that people often use `ns` and `in-ns` to move around namespaces, so they can't just use `(quit)` and expect it to work. (I hadn't recalled this during our previous discussion.)

Also there seems to be some strange new interaction with the `liner` work and how `ExitJoker()` exits, which I noticed when testing this. Specifically, terminal echoing is disabled after `(joker.os/exit 0)` or equivalent, but not after EOF, to exit the REPL. Submitted https://github.com/candid82/joker/issues/384.

But that's true for current `master`, so is not related to this PR. I hope to investigate it later, unless you get to it first!


jcburley:
Hmm, I guess we could document `(user/quit)` instead, which is shorter and yet likely to work in most circumstances?

candid82:
> This does the job, but illustrates the awkwardness of defining a namespaced function for this, in that people often use ns and in-ns to move around namespaces, so they can't just use (quit) and expect it to work. 

Hm... I think I would actually be OK with adding `(quit)` or `(exit)` to `joker.core`. That would align with the goal of making scripting easier as exiting is pretty common use case. Do you see any pitfalls with this (other than adding non-standard function to core namespace, which I think is justified in this case)?

I do like `exit` better than `quit`, but don't want it to conflict with `joker.os/exit`... Perhaps we could just remove `joker.os/exit`. It's a breaking change, but very easy for the users to accommodate...

jcburley:
> > This does the job, but illustrates the awkwardness of defining a namespaced function for this, in that people often use ns and in-ns to move around namespaces, so they can't just use (quit) and expect it to work.
> 
> Hm... I think I would actually be OK with adding `(quit)` or `(exit)` to `joker.core`. That would align with the goal of making scripting easier as exiting is pretty common use case. Do you see any pitfalls with this (other than adding non-standard function to core namespace, which I think is justified in this case)?
> 
> I do like `exit` better than `quit`, but don't want it to conflict with `joker.os/exit`... Perhaps we could just remove `joker.os/exit`. It's a breaking change, but very easy for the users to accommodate...

Interesting idea. `joker.os/exit` takes a single argument (the exit code), but a `joker.core/exit` or `joker.core/quit` could make that argument optional. `joker.os/exit` could then just call that, if we wanted to keep it around awhile.

jcburley:
I pinged borkdude on Slack in case he had any insights, because I wonder if we're missing something else about the keyword-vs-function issue, and he's another Clojure-variant implementor.

jcburley:
Per https://github.com/candid82/joker/pull/382#issuecomment-602793158, if we go this route (name-based, rather than keyword-based, exiting), I think being as much like `lein repl` would be great.

It looks like neither `quit` nor `exit` are actual functions:

```
$ lein repl
nREPL server started on port 49427 on host 127.0.0.1 - nrepl://127.0.0.1:49427
REPL-y 0.3.7, nREPL 0.2.12
Clojure 1.8.0
Java HotSpot(TM) 64-Bit Server VM 9.0.1+11
    Docs: (doc function-name-here)
          (find-doc "part-of-name-here")
  Source: (source function-name-here)
 Javadoc: (javadoc java-object-or-class-here)
    Exit: Control+D or (exit) or (quit)
 Results: Stored in vars *1, *2, *3, an exception in *e

user=> (identity +)
#object[clojure.core$_PLUS_ 0x7575e8c3 "clojure.core$_PLUS_@7575e8c3"]
user=> (identity quit)

CompilerException java.lang.RuntimeException: Unable to resolve symbol: quit in this context, compiling:(null:1:1) 
user=> (identity exit)

CompilerException java.lang.RuntimeException: Unable to resolve symbol: exit in this context, compiling:(null:1:1) 
user=> (exit 3)

CompilerException java.lang.RuntimeException: Unable to resolve symbol: exit in this context, compiling:(null:1:1) 
user=> quit hey there :foo "blah
  #_=> "
Bye for now!
$
```

(Similarly, `exit 3` ignores the excess arg and exits with a result code of 0.)

So it appears both `quit` and `exit` are treated as special forms, not  functions in `joker.core`, and only when they're entered "raw" or as function invocations without arguments.

Taking this approach would mean no changes to `joker.os/exit` are necessary.

This has become my favorite approach, though I still don't feel strongly about any particular approach. Though it special-cases recognizing these special forms, being in a REPL is something of a special case anyway; it avoids namespace issues; and it's consistent with a fairly popular (I would think) REPL for any Clojure variant.

What do you think?

jcburley:
Created https://github.com/candid82/joker/pull/386 as a speculative PR.

So you now have _three_ to choose from!

candid82:
This is addressed by https://github.com/candid82/joker/commit/08a3b0c94cc5dda72a38e4c86d142cf4cdb5ef01 (I went with `joker.core/exit` approach)

jcburley:
Great, thanks!

PR #386, opened 2020-03-23
Support 'quit' or 'exit' as special forms to exit REPL

jcburley:
This approximates `lein repl` better than https://github.com/candid82/joker/pull/382 or https://github.com/candid82/joker/pull/383.

It differs from `lein repl` in that it documents only the symbol forms (`quit` and `exit`), though (like `lein repl`) it supports the function forms (`(quit)` and `(exit)`) as well.


candid82:
I decided to go with joker.core/exit in the end. I personally think it's cleaner. The conflict with joker.os/exit doesn't seem to be causing any real issues, and I added zero arity to both joker.core/exit and joker.os/exit. https://github.com/candid82/joker/commit/08a3b0c94cc5dda72a38e4c86d142cf4cdb5ef01

jcburley:
Great, thanks!

PR #387, opened 2020-03-27
Add tags to usage docs

jcburley:
Putting tag (type) info in the generated docs seems useful, although many functions don't have it and probably should.

But I'm not thrilled with the resulting "look", because it is a bit cluttered. E.g.:

https://burleyarch.com/joker/docs/amd64-linux/joker.core.html#+

The tags seem to clutter up the args a bit. (I don't mind the return-type tag so much.)

One approach to mitigating that is to put commas between the arguments (i.e. a comma followed by a space). But that adds punctuation clutter even when not needed (when there are no tags).

So another approach is to add commas between the arguments only when any of the arguments has a tag.

A third approach is to not include the tags in the text, but support hover/mouse-over to show them (and maybe highlight the args that have such tags so the reader knows whether to bother hovering).

Fourth, maybe have a separate block (perhaps a second pane, horizontally speaking) showing the tagged versions of the usages.

Fifth, keep the presentation as-is and wait for people to complain and suggest alternatives.

Sixth, leave it as-is but don't generate tags by default, and have the script support a `--generate-tags` option that uses one of the above approaches, so people can try it out on their own systems.


candid82:
I agree it can be useful to see the types, although type annotations are certainly not complete and may not be 100% accurate. I also don't think the types should be shown by default. Perhaps show them on hover over like you proposed or have a separate "show types" link to the right of "source" link. Some experimentation is needed here...

jcburley:
I know just enough HTML to do a hover, but I think a "Show Types" link that pops up the extra info (somehow) is a great idea, though I don't know offhand how to do it in HTML.

I'll look into that and maybe give it a try. Once I figure out the HTML, changing `generate-docs.joke` to support it should be fairly straightforward. (Famous last words....)

jcburley:
@candid82 -- please check out the two prototypes I threw together here and let me know what you think:

https://burleyarch.com/joker/docs/amd64-linux/experiments.html

(They'd each need some fine-tuning, stylistically, at the very least.)

candid82:
@jcburley "Show Types" button doesn't work for me (did you forget to include javascript file?), but hover over solution looks pretty good!

jcburley:
> @jcburley "Show Types" button doesn't work for me (did you forget to include javascript file?), but hover over solution looks pretty good!

Weird -- it works for me, even on a different machine (my Mac laptop).

I'm trying to avoid JavaScript if possible -- so far, that seems to work. Both just uses css, though "Show Types" uses the separate CSS file, while Hover borrows a different approach by embedded the style in the HTML.

I wonder if there's some issue with your browser not picking up the **experiments.css** file in the same directory?

I thought "Show Types" might be better because it'd allow Copy-and-Paste-style capture of the block. But that doesn't work for me after all.

Hover is nice because it shows only the specific usage, and otherwise doesn't clutter the screen.

To make it really sing, though, I'd need to add appropriate code to highlight that Hover would actually work, so people would know to even try it. I suppose adding text saying something like "Hover over a usage to show types" would be really clear, but the usual approach is to put blue dots or something under the text, I gather?

Either way, the extra decoration would (I think) be emitted _only_ for usages (or usage blocks) that actually had type hints. Otherwise things would look exactly the same.

jcburley:
I've updated this PR to do the hover thing, but haven't fine-tuned the style (as that's not something I have much experience doing).

I've merged this into my `gostd` fork and regened and uploaded the resulting docs; see a sample here:

https://burleyarch.com/joker/docs/amd64-darwin/joker.core.html#comp

(You might have to do a full refresh to get the latest `main.css`.)

candid82:
I'll merge and tweak the styles later. Thanks!

candid82:
I decided to go with "show types" link after all: https://github.com/candid82/joker/commit/817716c405f4dfa886cb7c68c3aa7cb1fd85d7a5

jcburley:
> I decided to go with "show types" link after all: [817716c](https://github.com/candid82/joker/commit/817716c405f4dfa886cb7c68c3aa7cb1fd85d7a5)

Very nice!!

PR #389, opened 2020-03-29
Exclude joker.tgz from source control

jcburley:
As this is generated by `docs/dash.sh` and is a binary blob, yielding complaints on Mac OS X due to confusion over line endings, it seems like this shouldn't be in source control at all at this point. (On the `gostd` branch, it's per-architecture-and-OS anyway, as are the generated docs.)

Or, maybe tag it as a binary blob?


candid82:
It has to be in source control, that's how it's [distributed](https://github.com/candid82/joker/blob/master/docs/joker.xml#L3).
I [marked](https://github.com/candid82/joker/commit/8097a5d293c0b72b5f9d7e9f9ef2d3b0816c3fb0) is a binary.

jcburley:
> It has to be in source control, that's how it's [distributed](https://github.com/candid82/joker/blob/master/docs/joker.xml#L3).

Oh, right, of course!

> I [marked](https://github.com/candid82/joker/commit/8097a5d293c0b72b5f9d7e9f9ef2d3b0816c3fb0) is a binary.

Great, thanks!

PR #390, opened 2020-03-29
Fix joker.xml generation

jcburley:
Somehow I missed that the original approach wasn't working due to being run in `docs/dash/` instead of `docs/`.


candid82:
Thanks!

Issue #395, opened 2020-04-20
Ctrl-Z No Longer Works Since `liner` Adopted


jcburley:
Tracking this issue here:

https://github.com/peterh/liner/issues/131

Might it be appropriate to introduce a `(suspend)` function to `joker.core`, a la `(exit)`?

jcburley:
Related: https://github.com/candid82/joker/issues/402.

jcburley:
Fixed in 0.15.4 by dint of `liner` v1.4.0.

PR #396, opened 2020-05-04
Add joker.core/suspend and Support External SIGTSTP

jcburley:
This provides a workaround for https://github.com/candid82/joker/issues/395.

Known issues include:

* Joker's version of `liner` still does not support Ctrl-Z (generating a SIGTSTP)
* After an external SIGTSTP while at the prompt, the next line of input (following a SIGCONT) doesn't get added to history
* An external SIGTSTP while not at the prompt (e.g. while sleeping) leaves the terminal in the default state upon SIGCONT (e.g. arrow keys don't work)

These issues suggest that this implementation might not be the proper way to handle things and that `liner` itself should provide better (ideally, rock-solid) handling of SIGTSTP and SIGCONT than it currently does.

In particular, changing the `rl` variable in `repl.go`, i.e. the `State` instance, should be avoided. But `rl.Close()` seems to be the only way to restore the terminal settings before continuing the SIGTSTP with a SIGSTOP (and thus reaching the shell) and, though that doesn't seem to actually tear down the state, there doesn't seem to be an API to resume the same state (by re-establishing the earlier terminal settings).

If `liner` offered an `rl.Reopen()` API, such that the same `rl` (copy of `liner.State`) could be used across SIGTSTP/SIGCONT events, changing `repl.go` to use that, instead of instantiating a new `State` upon SIGCONT, might solve some of the issues above.

But for people (like myself) who frequently want to simply suspend a Joker session and resume it later, via a (synchronous) `(suspend)` invocation, this PR seems to do the trick fairly well.

### DIFF
--- a/core/common.go
+++ b/core/common.go
@@ -21,29 +21,28 @@ func OnExit(f func()) {
 
 type suspendCallbackFns struct {
 	leave func()
-	stop  func() // Only one gets called; nil's are ignored.
 	rtn   func()
 }
 
+var SuspendStopFn func()
 var suspendCallbacks []suspendCallbackFns
 
-func SuspendJoker() {
+func SuspendJoker() bool {
+	if SuspendStopFn == nil {
+		return false
+	}
 	for _, fns := range suspendCallbacks {
 		fns.leave()
 	}
-	for _, fns := range suspendCallbacks {
-		if fns.stop != nil {
-			fns.stop()
-			break
-		}
-	}
+	SuspendStopFn()
 	for _, fns := range suspendCallbacks {
 		fns.rtn()
 	}
+	return true
 }
 
-func OnSuspend(leave func(), stop func(), rtn func()) {
-	suspendCallbacks = append(suspendCallbacks, suspendCallbackFns{leave, stop, rtn})
+func OnSuspend(leave func(), rtn func()) {
+	suspendCallbacks = append(suspendCallbacks, suspendCallbackFns{leave, rtn})
 }
 
 func writeIndent(w io.Writer, n int) {
--- a/core/common.go
+++ b/core/common.go
@@ -19,6 +19,33 @@ func OnExit(f func()) {
 	exitCallbacks = append(exitCallbacks, f)
 }
 
+type suspendCallbackFns struct {
+	leave func()
+	stop  func() // Only one gets called; nil's are ignored.
+	rtn   func()
+}
+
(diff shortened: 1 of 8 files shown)


jcburley:
(Note that I'm happy to help out with Joker's fork of `liner`, if that'd be welcome.)

candid82:
Thank you for looking into this issue. I feel like it is better to handle it inside `liner`, and it looks like you think so as well. Since the author of `liner` doesn't want to add this to the canonical `liner`, I'll be happy to accept a PR in my fork.

candid82:
To clarify, I am not keen on adding `suspend` function to `joker.core`, but if CTRL+Z is handled by `liner` this would not be necessary, right?

jcburley:
> Thank you for looking into this issue. I feel like it is better to handle it inside `liner`, and it looks like you think so as well. Since the author of `liner` doesn't want to add this to the canonical `liner`, I'll be happy to accept a PR in my fork.

Great! I'll look into doing this soon, and agree it makes much more sense to simply handle it in `liner`, so Joker need know nothing about it, as was the case when it used `readline`.

jcburley:
> To clarify, I am not keen on adding `suspend` function to `joker.core`, but if CTRL+Z is handled by `liner` this would not be necessary, right?

Exactly!

jcburley:
FYI, I decided to go ahead and push what I had, even though it'll fail tests due to depending on new `liner` APIs that haven't yet been approved/merged.

I did this because a) the existing PR no longer seemed useful and b) it might be easier to assess my `liner` PR while looking at the new version of this one that depends on it.

jcburley:
Note that this PR is still not "clean". For diagnostic purposes, Ctrl-Z is echoed as `^Z [Joker]`, which should be changed to just echo `^Z`; and `go.mod` and `go.sum` are changed, which they should probably not be by this PR.

So really it's just a preview of what this new approach would look like in terms of impact on Joker.

jcburley:
This actually makes https://github.com/candid82/joker/issues/402 worse, in that Ctrl-C and Ctrl-Z stop working.

Withdrawing due to this and per discussion on https://github.com/candid82/liner/pull/1.

I'll research these issues further and see about submitting a PR for `liner` and/or a PR for `Joker`, as needed, to get Ctrl-Z, and proper line discipline across `SIGTSTP`/`SIGCONT` events, working for both REPL input and `(joker.core/readline)` within the REPL.

PR #403, opened 2020-05-10
Improve REPL Welcome Message

jcburley:
As Joker doesn't seem to need any changes to support Ctrl-Z (or Ctrl-\) properly, either in the REPL or when calling `(joker.core/read-line)` from it, these changes seem to be all that's left to be done, and they have no important dependency on any particular version of `liner`.


candid82:
Thanks!

PR #411, opened 2020-06-06
Avoid crash on (meta) with wrong # of arguments

jcburley:
Fixes https://github.com/candid82/joker/issues/410


candid82:
Thanks!

PR #414, opened 2020-07-14
Fix expected output to include my.test0

jcburley:


candid82:
Thanks! I'll add formatter-tests.sh to CircleCI build.

Issue #417, opened 2020-09-03
32bit ARM builds fail due to integer overflow in (generated) a_code.go


pmonks:
### Steps to reproduce:
```shell
$ git clone https://github.com/candid82/joker.git
Cloning into 'joker'...
remote: Enumerating objects: 9, done.
remote: Counting objects: 100% (9/9), done.
remote: Compressing objects: 100% (8/8), done.
remote: Total 13960 (delta 0), reused 3 (delta 0), pack-reused 13951
Receiving objects: 100% (13960/13960), 5.86 MiB | 10.78 MiB/s, done.
Resolving deltas: 100% (9413/9413), done.
$ cd joker
$ ./run.sh --version && ./build-arm.sh
v0.15.6
# github.com/candid82/joker/core
core/a_code.go:400057:2: constant 3242738688 overflows int
```

Couple of notes:
1. this works fine if `GOARCH=arm64` in `build-arm.sh`
2. ensuring joker works on 32bit arm is very low priority for me, since I mostly use a newer arm64-based Raspberry Pi these days (though I still have the old 32bit arm one lying around)

As a result, perhaps the fix here could simply be to change `GOARCH` to `arm64` in build-arm.sh?  That would certainly be a suitable solution IMHO.

jcburley:
Confirmed; this problem was introduced by the `gen_code` phase that optimizes startup time by building a "mini-Joker", aka `gen_code`, that then outputs its pertinent internal structures as new Go code, which (mostly) statically initializes large data structures that previously were dynamically initialized (mostly from internally stored Joker code or equivalent).

By building `gen_code` as a 32-bit executable, which the local `GOOS=linux` can execute, it'll output suitable values for the final 32-bit build of Joker.

PR #418, opened 2020-09-04
Use 32-bit binaries to generate code for 32-bit cross-build

jcburley:
Fixes https://github.com/candid82/joker/issues/417.


candid82:
Thanks!

Issue #419, opened 2020-09-10
linter segfault when `clojure.core/list` is excluded in a namespace containing a macro


benalbrecht:
joker crashes when running `joker --lint` against the following clojure source file:

```clojure
(ns foo
  (:refer-clojure :exclude [list]))

(defmacro bar [baz]
  `(identity ~baz))
```

stack trace:
```
$ joker --lint foo.clj 
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
        panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x74190d]

goroutine 1 [running]:
github.com/candid82/joker/core.TryParse.func1(0xc00056bb18)
        /Users/candid/go/src/github.com/candid82/joker/core/parse.go:1863 +0x16b
panic(0x923600, 0xe58640)
        /usr/local/Cellar/go/1.14/libexec/src/runtime/panic.go:967 +0x15d
github.com/candid82/joker/core.parseSymbol(0xb0ae60, 0xc0001d95f0, 0xc00056bbb8, 0x0, 0x0)
        /Users/candid/go/src/github.com/candid82/joker/core/parse.go:1808 +0x4cd
github.com/candid82/joker/core.Parse(0xb0ae60, 0xc0001d95f0, 0xc00056bbb8, 0xb0d200, 0xc000438980)
        /Users/candid/go/src/github.com/candid82/joker/core/parse.go:1834 +0x3f7
github.com/candid82/joker/core.parseList(0xb0a260, 0xc000438980, 0xc00056bbb8, 0x0, 0x0)
        /Users/candid/go/src/github.com/candid82/joker/core/parse.go:1643 +0x336
github.com/candid82/joker/core.Parse(0xb0a260, 0xc000438980, 0xc00056bbb8, 0x129aba0, 0xe52b00)
        /Users/candid/go/src/github.com/candid82/joker/core/parse.go:1832 +0x524
github.com/candid82/joker/core.parseSeq(0xb0d200, 0xc000438b00, 0xc00056bbb8, 0x0, 0x0, 0xc000567210)
        /Users/candid/go/src/github.com/candid82/joker/core/parse.go:625 +0xfa
github.com/candid82/joker/core.parseList(0xb0a260, 0xc000438b40, 0xc00056bbb8, 0x0, 0x0)
        /Users/candid/go/src/github.com/candid82/joker/core/parse.go:1664 +0x5de
[...]

PR #420, opened 2020-09-10
Fix crash when linting file that excludes 'list' from namespace with …

jcburley:
…macro

Fixes: https://github.com/candid82/joker/issues/419


candid82:
Thanks!

PR #421, opened 2020-09-11
Fix broken doc links in DEVELOPER.md

dpritchett:
New Joker fan here! Been having a great time migrating all my old bash junk to joker this week 😍 

The current links out of DEVELOPER to LIBRARIES and README 404 when reading the developer instructions from GitHub.

The new ones use relative links instead of absolute links; they now resolves to the actual hosted page at e.g. https://github.com/candid82/joker/blob/master/LIBRARIES.md and should also work within branches if browsing non-`master` branches.


candid82:
Thanks!

Issue #423, opened 2020-09-28
##Inf, ##-Inf, and ##NaN are not recognised


Briaoeuidhtns:
When linting or at a repl, joker does not recognise `##Inf`, `##-Inf`, or `##NaN`.

For instance, linting a file containing
```clojure
(print ##Inf)
```
gives
```
test.clj:1:12: Read warning: No reader function for tag Inf
test.clj:1:13: Read error: Unexpected )
```

PR #424, opened 2020-09-28
Support ##Inf, ##-Inf, and ##NaN

jcburley:
Fixes: https://github.com/candid82/joker/issues/423


candid82:
Looks great, thanks a lot @jcburley!

PR #430, opened 2020-12-24
Rename/refactor the Ensure/Assert functions

jcburley:
Currently the caller of `Assert*` is given a binary choice: pass an empty message, letting the API (if necessary) compose a useful one with the expected and actual type information; or, provide a complete message, requiring the caller to go to the expense of doing that even if (as is most often the case) it's not needed due to the type being in conformance.

Herein, the message is converted to a pattern (the default being "`%s`"), and the pattern is expanded only when needed, providing the current default message as the only argument to the pattern. (For most callers, this is sufficient; a few must still precompute a pattern based on runtime info, so if that is deemed an issue, arbitrary additional arguments could be supported via e.g. "`...interface{}`" and such. Callers that don't want the default argument inserted at all can and should put "`%.s`", which indicates a 0-length-truncated string, somewhere in their pattern.)

The result is simplified, as well as slightly faster, code. The resulting messages are different , but that -- surprisingly -- doesn't seem to affect the current set of tests:

```
$ joker
Welcome to joker v0.15.7. Use '(exit)', EOF (Ctrl-D), or SIGINT (Ctrl-C) to exit.
user=> (double "hey")
<joker.core>:2040:23: Eval error: Cannot cast "hey" (type: String) to Double
Stacktrace:
  global <repl>:1:1
  core/double <joker.core>:2040:23
user=> ^D
$ ./joker
Welcome to joker v0.15.7. Use '(exit)', EOF (Ctrl-D), or SIGINT (Ctrl-C) to exit.
user=> (double "hey")
<joker.core>:2040:23: Eval error: Cannot cast "hey": Expected Number, got String
Stacktrace:
  global <repl>:1:1
  core/double <joker.core>:2040:23
user=> ^D
$
```

Also, I've consistently been confused by the distinction between "Assert" and "Ensure". While I don't have a strong feeling about which term is used, it seems like only one should be (and "Assert" implies, to me, a mere check, not also the return of a value of the desired datatype).

So I've renamed "`Assert<T>`" to "`EnsureObjectIs<T>`", after first renaming "`Ensure<T>`" to "`EnsureArgIs<T>`" (where appropriate). I think this'll make it a little easier to read code without resorting to looking at `types_assert_gen.go` so frequently.

### DIFF
--- a/DEVELOPER.md
+++ b/DEVELOPER.md
@@ -360,7 +360,7 @@ initialized to a global Clojure namespace with `NSFULLNAME`
 to the set `joker.core/*loaded-libs*`:
 
 ```Go
-var fooNamespace = GLOBAL_ENV.EnsureLib(MakeSymbol("joker.foo"))
+var fooNamespace = GLOBAL_ENV.EnsureSymbolIsLib(MakeSymbol("joker.foo"))
 ```
 
 `a_foo.go` finishes with:
--- a/core/channel.go
+++ b/core/channel.go
@@ -51,7 +51,7 @@ func MakeChannel(ch chan FutureResult) *Channel {
 }
 
 func ExtractChannel(args []Object, index int) *Channel {
-	return EnsureChannel(args, index)
+	return EnsureArgIsChannel(args, index)
 }
 
 func (ch *Channel) Close() {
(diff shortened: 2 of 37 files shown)


candid82:
This is a nice improvement. Thanks a lot!

PR #431, opened 2021-01-08
Fix (compare true 2) diagnostic message

jcburley:
My most-recent PR missed changing this one, so broke the diagnostic.


candid82:
Thanks!

PR #432, opened 2021-01-08
Further simplify ensurers

jcburley:
This moves more of the repetitive code into a couple of new functions.

As a compiler writer, I'd expect the resulting binaries to be smaller; but they seem to get a tiny bit larger with this changeset. I'm not sure why, unless Go is somehow better at consolidating repeating code, when it's beyond a certain size or complexity, than writing (generating) less of it in the first place, when it might be not bothering?

But I think the results are easier to look at and understand, and (if merged) I think it puts my next chunk of `gostd` work in a better position, as that needs another "dimension" (or perhaps several) of ways to extract objects.

### DIFF
--- a/core/gen/gen_types.go
+++ b/core/gen/gen_types.go
@@ -12,7 +12,6 @@ type (
 		Name     string
 		TypeName string
 		ShowName string
-		NilValue string
 	}
 )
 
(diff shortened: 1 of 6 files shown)


core/gen/gen_types.go
```diff
@@ -22,34 +22,26 @@ package core
...
+	if c, yes := obj.({{.TypeName}}); yes {
 		return c
-	default:
-		if pattern == "" {
-			pattern = "%s"
-		}
-		msg := fmt.Sprintf("Expected %s, got %s", "{{.ShowName}}", obj.GetType().ToString(false))
-		panic(RT.NewError(fmt.Sprintf(pattern, msg)))
 	}
+	panic(FailObject(obj, "{{.TypeName}}", pattern))
```
candid82, line 34: Should not we pass `"{{.ShowName}}"` here instead of `"{{.TypeName}}"` to preserve current behavior?

jcburley: Ah, indeed, I would think so! Will confirm and update as appropriate.

candid82:
Thanks!

Issue #434, opened 2021-02-18
Wrong format result.


glyh:
I use joker to format my clojure code. 

When I have some small double values, say 1e-7, the formatter just output 0.0.

This would cause some fault on Clojure Codes, please fix it.

jcburley:
I'm thinking this is more than a formatting issue:

```
$ joker
Welcome to joker v0.15.7. Use '(exit)', EOF (Ctrl-D), or SIGINT (Ctrl-C) to exit.
user=> (println 1e-9)
0.0
nil
user=> (println 1e-7)
0.0
nil
user=> (println 1e-5)
0.00001
nil
user=> ^D
$
```

Gonna take a quick look at the code myself, in case something obvious pops out.

jcburley:
More likely a printing than a reading issue:

```
user=> (< 1e-9 1e-7)
true
user=> (< 1e-9 1e-9)
false
user=> (< 1e-9 1e-11)
false
user=> 
```

jcburley:
Also, 1.9999999 renders as "2.0", but is stored such that it compares as less than 2.0.

jcburley:
I've got a fix ready for a PR, just adding tests.

PR #435, opened 2021-02-18
Fix float output

jcburley:
Fixes https://github.com/candid82/joker/issues/434.

Note that Go's "%g" format specifier always outputs exactly three non-whitespace runes after the "e" (for exponent), besides other differences to Clojure/JVM, such as using "e" instead of "E", and not ensuring the string preceding "e"/"E" has a decimal point and at least one trailing digit.

So `1e-7` renders as "1e-07", for example. (Clojure 1.10.1 renders this "1.0E-7".)

ClojureScript's output looks more like what Joker would with this PR applied, except it renders `1e-7` (or any numerical equivalent) as "1e-7", omitting the unnecessary leading 0. (It also renders values that can be precisely converted to integers as integers; this PR avoids doing that via some extra code, which I think is necessary to avoid bugs such as when formatting code that might depend on having the types strictly correct. In ClojureScript, or at least lumo 1.10.1, integers and floats all appear to become `Number` type; that's not true of Clojure nor Joker.)

As the Clojureverse seems to generally tend towards relying on the underlying run-time environment (JVM, JS, Go, etc), rather than strict specifications, to make such decisions, it should be okay for Joker to rely on "%g" with the current extra code to ensure the result would be read back in as a `Double`; but if we want to go further and remove the leading "0" after "e-", let me know and I'll make that change. Ditto replacing "e" with "E" and/or ensuring the preceding text looks like a floating-point value in isolation (e.g. "1.0" instead of just "1" before an "e"/"E").

### DIFF
--- a/core/object.go
+++ b/core/object.go
@@ -1086,14 +1086,7 @@ func (d Double) ToString(escape bool) string {
 	if math.IsNaN(dbl) {
 		return "##NaN"
 	}
-	res := fmt.Sprintf("%f", dbl)
-	var i int
-	for i = len(res) - 1; res[i] == '0'; i-- {
-	}
-	if res[i] == '.' {
-		return res[0 : i+2]
-	}
-	return res[0 : i+1]
+	return fmt.Sprintf("%g", dbl)
 }
 
 func (d Double) Equals(other interface{}) bool {
--- a/core/object.go
+++ b/core/object.go
@@ -1086,7 +1086,11 @@ func (d Double) ToString(escape bool) string {
 	if math.IsNaN(dbl) {
 		return "##NaN"
 	}
-	return fmt.Sprintf("%g", dbl)
+	res := fmt.Sprintf("%g", dbl)
+	if strings.ContainsAny(res, ".e") {
+		return res
+	}
+	return res + ".0"
 }
 
 func (d Double) Equals(other interface{}) bool {
--- a/tests/eval/printer.joke
+++ b/tests/eval/printer.joke
@@ -16,3 +16,11 @@
     "##Inf" (joker.math/inf 1)
     "##-Inf" (joker.math/inf -1)
     "##NaN" (joker.math/nan)))
+
+(deftest print-double-values
+  (are [s v] (= s (pr-str v))
+    "1.9999999" 1.9999999
+    "1e-09" 1e-9
+    "1e-07" 1e-7
+    "0.001" 1e-3
+    "1.0" 1e0))


candid82:
Thanks a lot! I think the way you implemented it is fine as is. I do wonder why I [changed](https://github.com/candid82/joker/commit/05f822a85afeb11c468e5363a6fd1999b8a93fd6#diff-f811ad106ed3533deeb7e92d73931fd62118c28a230b57d5cf6d30e1bd278113L1072) the original implementation that was based on using `%g`... It did not handle `e`, but a simple tweak (like in your solution) would have handled this... Perhaps I did not want numbers like `0.00001` to be formatted / printed as `1e-05` (as it would do with `%g`)... But it's probably OK. Certainly better than to lose precision as with existing implementation.

jcburley:
> Thanks a lot! I think the way you implemented it is fine as is. I do wonder why I [changed](https://github.com/candid82/joker/commit/05f822a85afeb11c468e5363a6fd1999b8a93fd6#diff-f811ad106ed3533deeb7e92d73931fd62118c28a230b57d5cf6d30e1bd278113L1072) the original implementation that was based on using `%g`... It did not handle `e`, but a simple tweak (like in your solution) would have handled this... Perhaps I did not want numbers like `0.00001` to be formatted / printed as `1e-05` (as it would do with `%g`)... But it's probably OK. Certainly better than to lose precision as with existing implementation.

Thanks! I've since had a few ideas concerning this; let me know whether you'd like me to follow up on any of them:

- The new tests should be format-agnostic, by verifying only that a given value, when output and read back in, produces the same value (rather than the output matches a specific string).
- More width could be specified, to allow `0.00001` to be output as such.
- The format specifier/function could be made dynamically adjustable in some fashion, for users who want that flexibility. I suspect use of dynamic variables might be too slow (to look up each time a number is to be formatted); but static variables modifiable via (private?) functions in `joker.core` shouldn't.
- As Joker is now useful for formatting Clojureverse code (and there are a variety of formats in effect among Clojure implementations), besides the utility of dynamic setting of the format specifier, it might be helpful to think of the default format as whatever is closest to what people would normally type in their Clojure source code. Towards this end, I think few people would type "1e-07", but would type "1e-7" instead, i.e. they'd not specify the leading "0".

candid82:
Those are all interesting ideas. But I think we should separate two use cases here:
1. Printing doubles when using Joker as a formatter.
2. Printing doubles in regular mode.

For #2, the current implementation seems more than adequate to me. If people need more control over how doubles are printed, they can always use `format` function which gives you access to the full power lf Go's `fmt.Sprintf`.

For #1, I think ultimately the right way to do it is to capture the original string representation of a double as it was presented in the source code. The formatter should preserve the original formatting of doubles (and other literals for that matter). This seems rather involved change so I am not 100% the benefits outweigh the cost... So far I've been under impression I was the only one who used Joker as a formatter and the occasional re-formatting of some literals has not been a big problem for me. 

To address the ideas above:
1. I'd prefer to check the actual string output. If we  change the formatting again (intentionally or not), we'll want the tests to show that.
2. Is it possible to make `%g` print `0.00001` without converting it into `1e-05`?
3. I don't think it's worth the effort. As mentioned above, people can always use `format` for more control.
4. I guess that would be an alternative approach to the one I described above (preserving the original literal representation). I just think any specific way of formatting literals will never satisfy everyone.

jcburley:
> Those are all interesting ideas. But I think we should separate two use cases here:
> 
> 1. Printing doubles when using Joker as a formatter.
> 2. Printing doubles in regular mode.

That makes sense.

> For #2, the current implementation seems more than adequate to me. If people need more control over how doubles are printed, they can always use `format` function which gives you access to the full power lf Go's `fmt.Sprintf`.

That seems reasonable. Only those who have trouble adapting their code to use their own function, rather than relying on low-level print-like calls, to render doubles, would likely have problems here; maybe a few others might hit performance problems for sufficiently high volumes of output (as their own Joker-coded functions would not likely run as fast as the hand-coded Go that is normally used). In both cases, it seems reasonable to wait until somebody in one of those two groups makes a good case with real code.

> For #1, I think ultimately the right way to do it is to capture the original string representation of a double as it was presented in the source code. The formatter should preserve the original formatting of doubles (and other literals for that matter). This seems rather involved change so I am not 100% the benefits outweigh the cost... So far I've been under impression I was the only one who used Joker as a formatter and the occasional re-formatting of some literals has not been a big problem for me.

I was thinking along the same lines (preserving the original representation and using that where available). If you'd like me to take a swing at that, let me know; smaller-scoped projects like that (and this one) often fit into my schedule these days.

> To address the ideas above:
> 
> 1. I'd prefer to check the actual string output. If we  change the formatting again (intentionally or not), we'll want the tests to show that.

Ok. Still, it might make sense to add a (larger?) set of tests that just check for identity.

> 2. Is it possible to make `%g` print `0.00001` without converting it into `1e-05`?

Apparently not, based on the experimentation I just did using `(format ...)`! Seems strange that the point at which such conversion happens isn't configurable; I'm pretty sure it is in some other language environments (such as Fortran). Perhaps the thinking is that people would rather not have to count more than some number (here, three) of zeros after a decimal point? But they're okay with as many as five zeros before one?

> 3. I don't think it's worth the effort. As mentioned above, people can always use `format` for more control.

Ok.

> 4. I guess that would be an alternative approach to the one I described above (preserving the original literal representation). I just think any specific way of formatting literals will never satisfy everyone.

You're probably right.

Given that fact, I agree it makes sense to leave things as they are (since merging this PR), other than perhaps to add another set of tests for identity instead of formatting, because the current approach represents the least-possible "massaging" of the rendered output while preserving the precise numeric value and also while ensuring that, when the rendered/printed form is read back in, the result has the same type as the original.

candid82:
Yes, I think additional tests for identity won't hurt. 
Feel free to take a stab at preserving the original representation for literals (at least in formatter mode) if you'd like.
And as usual, thank you very much for your contributions both in form of code and thoughtful discussions!

PR #437, opened 2021-02-22
Add format-independent tests ensuring sample doubles are rendered and…

jcburley:
… read back in the same way.


candid82:
Thanks!

Issue #436, opened 2021-02-20
ClojureScript Dialect Treats ".001" as Number, Joker Doesn't


jcburley:
In ClojureScript (via lumo), a floating-point literal beginning with "." is parsed as a Number, not as a Symbol (which is how Joker and Clojure itself parse it).

Assuming this isn't a bug in ClojureScript (that might get fixed soon), perhaps Joker should do the same when `--dialect cljs` is in effect?

E.g. consider:

```
$ joker --lint --dialect cljs --file -
(def .001 "hey")
$ joker --lint --dialect cljs --file -
(def 1 "hey")
<stdin>:1:6: Parse error: First argument to def must be a Symbol
$
```

The first line is not accepted by lumo:

```
$ lumo
Lumo 1.10.1
ClojureScript 1.10.520
Node.js v11.13.0
 Docs: (doc function-name-here)
       (find-doc "part-of-name-here")
 Source: (source function-name-here)
 Exit: Control+D or :cljs/quit or exit

cljs.user=> (def .001 "hey")
            ⬆
WARNING: No such namespace: , could not locate .cljs, .cljc, or JavaScript source providing "" at line 1 
            ⬆
WARNING: Use of undeclared Var /001 at line 1 
            ⬆
WARNING: No such namespace: , could not locate .cljs, .cljc, or JavaScript source providing "" at line 1 
            ⬆
WARNING: Use of undeclared Var /001 at line 1 
Invalid left-hand side in assignment
	 (new)
	 createScript (vm.cljs:266:10)
	 Object.runInThisContext (vm.cljs:314:10)
	 (Object.It)
	 (Object.lumo.repl.caching_node_eval)
	 (NO_SOURCE_FILE <embedded>:6029:9)
	 z (NO_SOURCE_FILE <embedded>:6030:22)
	 (NO_SOURCE_FILE <embedded>:6025:264)
	 Function.cljs.core.trampoline.cljs$core$IFn$_invoke$arity$1 (NO_SOURCE_FILE <embedded>:1927:142)
	 Function.cljs.core.trampoline.cljs$core$IFn$_invoke$arity$variadic (NO_SOURCE_FILE <embedded>:1927:280)
```

(Joker's diagnostics sure seem a lot...simpler and clearer, at least here!)

candid82:
Agreed, Joker should read `.001` as a number if dialect is `cljs`. I'll look into this unless you want to.

PR #438, opened 2021-02-22
ClojureScript dialect: Treat leading .<digit> as number

jcburley:
Fixes https://github.com/candid82/joker/issues/436, though I'm not convinced this is intentional behavior in ClojureScript.


candid82:
👍 
Even if it was not intentional behavior, it's very unlikely they are going to change it as it would break compatibility.

PR #439, opened 2021-02-22
Preserve floating-point literals

jcburley:
This handles only normal floating-point (`Double`) values, as a proof of concept.

The performance actually seems to improve when formatting a large file full of floating-point values. I guess the overhead of preserving the original string is less than that of rendering an arbitrary floating-point value (versus just printing that original string, when available)?

Other numeric types can be done almost (but not quite) as easily. They all need to either preserve, or reconstitute, the entire string (which the `Double` case already did naturally).

Beyond that, would it make sense to handle other literals (strings, symbols, keywords, etc) similarly? That might not be as obviously desirable as preserving e.g. `0100` (versus the current behavior of formatting it as `64`).

Let me know what you think about this initial attempt. Seems like a useful feature (if not expected behavior) when formatting code!


candid82:
This looks good! Yes, it makes sense to do the same for all numeric types. I cannot think of any possible variations in formatting symbols and keywords. And strings should already be covered by [this](https://github.com/candid82/joker/blob/master/core/read.go#L514). So I think it only makes sense to preserve originals for numbers, unless you can think of other use cases?

jcburley:
I think the numerics are indeed all that require such preservation!

PR #440, opened 2021-02-22
Preserve numeric literals

jcburley:
Finish the work of preserving all numeric literals when formatting. An added bonus is that the code that reads numbers actually gets measurably faster, because it's no longer necessary to precompute an `error` object in case parsing fails, as the original string is now always at hand.


candid82:
Looks great, thank you!

Issue #18, opened 2017-08-01
Wrong number of args (2) passed to #'joker.core/array


eneroth:
I spotted this in the [atom-parinfer](https://github.com/oakmac/atom-parinfer) source code. I suspect that it's guessing wrongly as to what `array` is being referenced in the code.

![screen shot 2017-08-01 at 08 17 58](https://user-images.githubusercontent.com/333974/28809129-55664152-7692-11e7-8188-3d8367f3f06b.png)

candid82:
Fixed by https://github.com/candid82/joker/commit/19129872c2eda90d5bdbb9ae5e24fcf1a059f5b6

PR #445, opened 2021-02-25
Linter Support for Validating Symbol/Keyword Characters

jcburley:
This is one approach to addressing https://github.com/candid82/joker/discussions/444.

Note that it changes Joker behavior in that it (by default) disallows em dashes, glyphs/graphics, and other such things in symbol/keyword names.

While that would make Joker less compatible with Clojure, I think it'd make a better argument for Joker as an effective linting and formatting tool, as it seems (to me) very unlikely any developers actually want such characters in symbol or keyword names in their source code.

But that default could be easily changed (to `--valid-letters unicode`) if stricter compatibility, out of the box, is desired.

If `--valid-letters ascii` is considered useful to/by some users, it might make sense to add a `--valid-digits` option to specify either `unicode` (the default) or `ascii` (only `[0-9]`); that possibility weighed on the decision on how to name `--valid-letters`. Limiting valid digits to only ASCII would affect parsing of not only symbol and keyword names, but (I would think) of numbers as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,24 +139,6 @@ test.clj:1:1: Parse warning: let form with empty body
 ```
 The output format is as follows: `<filename>:<line>:<column>: <issue type>: <message>`, where `<issue type>` can be `Read error`, `Parse error`, `Parse warning` or `Exception`.
 
-### Valid Symbol (and Keyword) Characters
-
-The `--valid-letters` option specifies valid characters for symbol and keyword "letters", which are the non-special, non-digit characters that make up symbol names.
-
-While Clojure 1.10 allows, for example, em dashes in symbol names (which can lead to confusion), Joker does not; by default, it allows only Unicode letters (category L). E.g. `(def é "hey")` works as expected. `--valid-letters letters` specifies this explicitly.
-
-Earlier versions of Joker allowed any Unicode code points beyond MaxLatin1 (255, aka `0xff`) in addition to the ASCII letters (`[A-Z][a-z]`). This is more consistent with Clojure, in that it allows em dashes; e.g. `(def a–b "wow")`. Specify `--valid-letters unicode` for this behavior.
-
-Some developers prefer "strict ASCII". Joker supports this for symbol letters (though not for digits, which may be any Unicode digits) via `--valid-letters ascii`.
-
-On the other hand, `--valid-letters any` allows any character (not otherwise special, such as delimeters), including ASCII control characters.
-
-Note that `--valid-letters` affects only how symbols are parsed; `joker.core/symbol` is able to form a symbol from any string, regardless of content.
-
-Also note that, though most closely associated with linting, `--valid-letters` governs how symbols are parsed regardless of mode of operation (linting, running scripts, formatting, etc).
-
-Finally, the environment variable `JOKER_VALID_LETTERS` may be used to specify the default for `--valid-letters`.
-
 ### Integration with editors
 
 - Emacs: [flycheck syntax checker](https://github.com/candid82/flycheck-joker)
@@ -251,8 +233,8 @@ When linting directories Joker lints all files with the extension corresponding 
 When linting directories Joker can report globally unused namespaces and public vars. This is turned off by default but can be enabled with `--report-globally-unused` flag, e.g. `joker --lint --working-dir my-project --report-globally-unused`. This is useful for finding "dead" code. Some namespaces or vars are intended to be used by external systems (e.g. public API of a library or main function of a program). To exclude such namespaces and vars from being reported as globally unused list them in `:entry-points` vector in `.joker` file, which may contain the names of namespaces or fully qualified names of vars. For example:
 
 ```clojure
-:entry-points [my-project.public-api
-               my-project.core/-main]
+{:entry-points [my-project.public-api
+                my-project.core/-main]}
 ```
 
 ### Optional rules
@@ -260,13 +242,13 @@ When linting directories Joker can report globally unused namespaces and public 
 Joker supports a few configurable linting rules. To turn them on or off set their values to `true` or `false` in `:rules` map in `.joker` file. For example:
 
 ```clojure
-:rules {:if-without-else true
-        :no-forms-threading false}
+{:rules {:if-without-else true
+         :no-forms-threading false}}
 ```
 
 Below is the list of all configurable rules.
 
-|          Rule          |                      Description                      | Default value |
+| Rule                   | Description                                           | Default value |
 |------------------------|-------------------------------------------------------|---------------|
 | `if-without-else`      | warn on `if` without the `else` branch                | `false`       |
 | `no-forms-threading`   | warn on threading macros with no forms, i.e. `(-> a)` | `true`        |
@@ -274,6 +256,10 @@ Below is the list of all configurable rules.
 | `unused-keys`          | warn on unused `:keys`, `:strs`, and `:syms` bindings | `true`        |
 | `unused-fn-parameters` | warn on unused fn parameters                          | `false`       |
 | `fn-with-empty-body`   | warn on fn form with empty body                       | `true`        |
+| `invalid-letters`      | warn on symbols/keywords with non-special/non-digit   | `:letters`    |
+|                        | characters (normally letters) that are not `:unicode` |               |
+|                        | code points or letters; not `:letters` (category L);  |               |
+|                        | or not `:ascii` letters (`[A-Z][a-z]` as a regexp)    |               |
 
 Note that `unused binding` and `unused parameter` warnings are suppressed for names starting with underscore.
 
--- a/README.md
+++ b/README.md
@@ -256,7 +256,7 @@ Below is the list of all configurable rules.
 | `unused-keys`          | warn on unused `:keys`, `:strs`, and `:syms` bindings | `true`        |
 | `unused-fn-parameters` | warn on unused fn parameters                          | `false`       |
 | `fn-with-empty-body`   | warn on fn form with empty body                       | `true`        |
-| `invalid-letters`      | warn on symbols/keywords with non-special/non-digit   | `:letters`    |
+| `invalid-letters`      | warn on symbols/keywords with non-special/non-digit   | `:unicode`    |
 |                        | characters (normally letters) that are not `:unicode` |               |
 |                        | code points or letters; not `:letters` (category L);  |               |
 |                        | or not `:ascii` letters (`[A-Z][a-z]` as a regexp)    |               |
--- a/README.md
+++ b/README.md
@@ -256,7 +256,7 @@ Below is the list of all configurable rules.
 | `unused-keys`          | warn on unused `:keys`, `:strs`, and `:syms` bindings | `true`        |
 | `unused-fn-parameters` | warn on unused fn parameters                          | `false`       |
 | `fn-with-empty-body`   | warn on fn form with empty body                       | `true`        |
-| `invalid-letters`      | warn on symbols/keywords with non-special/non-digit   | `:unicode`    |
+| `valid-letters`        | warn on symbols/keywords with non-special/non-digit   | `:unicode`    |
 |                        | characters (normally letters) that are not `:unicode` |               |
 |                        | code points or letters; not `:letters` (category L);  |               |
 |                        | or not `:ascii` letters (`[A-Z][a-z]` as a regexp)    |               |
--- a/README.md
+++ b/README.md
@@ -139,6 +139,24 @@ test.clj:1:1: Parse warning: let form with empty body
 ```
 The output format is as follows: `<filename>:<line>:<column>: <issue type>: <message>`, where `<issue type>` can be `Read error`, `Parse error`, `Parse warning` or `Exception`.
 
+### Valid Symbol (and Keyword) Characters
+
+The `--valid-letters` option specifies valid characters for symbol and keyword "letters", which are the non-special, non-digit characters that make up symbol names.
+
+While Clojure 1.10 allows, for example, em dashes in symbol names (which can lead to confusion), Joker does not; by default, it allows only Unicode letters (category L). E.g. `(def é "hey")` works as expected. `--valid-letters letters` specifies this explicitly.
+
+Earlier versions of Joker allowed any Unicode code points beyond MaxLatin1 (255, aka `0xff`) in addition to the ASCII letters (`[A-Z][a-z]`). This is more consistent with Clojure, in that it allows em dashes; e.g. `(def a–b "wow")`. Specify `--valid-letters unicode` for this behavior.
+
+Some developers prefer "strict ASCII". Joker supports this for symbol letters (though not for digits, which may be any Unicode digits) via `--valid-letters ascii`.
+
+On the other hand, `--valid-letters any` allows any character (not otherwise special, such as delimeters), including ASCII control characters.
+
+Note that `--valid-letters` affects only how symbols are parsed; `joker.core/symbol` is able to form a symbol from any string, regardless of content.
+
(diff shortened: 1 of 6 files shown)


candid82:
I like the idea in general, but I am hesitant to change Joker's runtime behavior. What about making this change strictly for the linter mode? In this case `valid-letters` option would go into `.joker` config file (rather than being a command line argument). For linting purposes this would be more flexible as you can create project specific `.joker` file with the desired `valid-letters` option. Also, `any` option probably would not make sense as it would allow invalid (from Clojure perspective) symbols and keywords.

jcburley:
I just pushed a commit that unwinds the option and env var stuff, leaves the old implementation in for now, fixes a couple of problems in `README.md`, and therein proposes a new `.joker` config that'd deal with this.

Please let me know what you think of that new config; it's a bit different from the others (takes a keyword rather than a boolean).

Once we nail down what it should look like, I'll work on implementing it (I guess as a warning, rather than the current implementation, which simply rejects problematic characters), unless you want to go at it first (since you're more aware of how people use Joker as a linter than I).

candid82:
The proposed config option looks good. And yes, it should be a warning, not an error (another argument for making it a linter mode only check). Feel free to proceed with the implementation.

jcburley:
I've got the above basically ready to go, but the more I look into this, the less I like this specific approach. Turns out:

- Clojure accepts pretty much any non-special character as a valid symbol char, even as a valid initial char
- ClojureScript (that is, lumo) does too, though no-break-space (0xa0) is treated as a space
- Joker thus stands alone by disallowing the entire Latin1 (ASCII 0x80-0xFF) set as well as most control characters (including 0x00 and others)

I find Clojure and ClojureScript's behavior rather bizarre, but in light of it, I'm not convinced there's any utility in the `:unicode` setting agreed-upon above (in my proposed updated PR and your approval) after all.

What I now think are the "interesting" settings (along with renaming the `:rules` key for simplicity) would be:

- `:valid-letters :any`, allowing anything the (Joker) parser allows
- `:valid-letters :printables`, allowing `unicode.IsPrint(r)`
- `:valid-letters :ascii`, allowing only `r <= unicode.MaxASCII`

Further, the Joker parser should probably be changed to be as permissible as Clojure/ClojureScript (as we've done recently with a leading `.`).

An argument could be made for allowing `unicode.IsLetter(r) || unicode.IsDigit(r)`, but I'm not sure who would want that, how we'd name it, and why they wouldn't also want the option of also allowing `unicode.IsPunct(r)`, `unicode.IsSymbol(r)`, maybe others.

So I think the main things to offer (besides changing Joker to parse more like Clojure and ClojureScript do) are summed up as:

- Allow whatever the parser allows
- Allow only "printable" characters (no non-breaking spaces, control characters, and the like)
- Allow only the original ASCII (7-bit) set, basically `[a-zA-Z0-9]`

For linting, it seems likely that most users would want the second option (printables-only), if not the third, as the default; I don't know.

I'm about to push what I have, in case it's useful for evaluation at least; but, given what I've learned, I don't think it should be Pulled, nevermind released as-is, unless we're okay with "breaking" it (e.g. disallowing `:unicode` and `:letters`) later on.

jcburley:
Also please note the discussion here:

https://github.com/clj-kondo/clj-kondo/issues/1179

In light of that, the `:rules` keyword defined herein might be worth renaming to something like `:valid-ident` and perhaps we could also support a `:valid-edn true` rule (if we can truly encompass whatever would be involved in that; I don't know).

jcburley:
It also stands to reason some might appreciate a `:valid-string` rule that ensures strings are composed of printable (even just ASCII) characters.

Generally, while the flexibility of specifying arbitrary Unicode characters in input/code files is often useful, for source code it can be very confusing (potentially leading to bugs) to have non-printing characters (other than the expected traditional whitespace) in it, in lieu of explicit coding using backslashes and such.

jcburley:
If you're interested in changing Joker to parse more like how Clojure does, I've got the beginnings of a PR that would do that. Though I haven't studied ClojureScript source code, I have reviewed Clojure's parsing code (LispReader.java) and attempted to better replicate it in a branch.

Note that, since Joker defines its own language, it _could_ define its own parse rules. E.g. it could keep the same (or similar, if we want to tweak them) rules for its own language, while relaxing them for Clojure/ClojureScript parsing.

Towards that end, it's worth considering whether to widen the scope of `--dialect` and the like so it encompasses all processing, not just linting. Right now, my experimental branch applies Joker's current parse rules only when linting with `DIALECT == JOKER` in effect; reading/evaluating Joker code is subject to the new, Clojure-like, relaxed parsing, in which any character (rune) that doesn't obviously start some other kind of token starts a symbol.

candid82:
I think it'd be good to change Joker's symbol/keyword parsing to match Clojure behavior. As I understand this will not break any existing code as it allows a superset of symbols/keywords compared to what's currently allowed.

I am OK with changing the option name to `:valid-ident` and I like the three options you proposed: `:any`, `:printable`, `:ascii`. The default could be `:ascii` as that's what most developers would want and this would allow to solve the original problem: warn on em dash.

What do you think we should do for those who do want larger Unicode set but still want to avoid em dashes and such? One solution could be adding the forth option: an arbitrary regex (for an individual symbol's character)... But that may be pretty annoying to deal with (you would have to know all Unicode classes that you want to allow etc). Another solution would be to add a separate option like `:ident-blacklist` that would list the characters that are explicitly forbidden.

Not sure about `:valid-edn`. I'd rather not add anything without specific complaint / use case, and I am not 100% sure what problem that option would solve. The same goes for `:valid-string`. I feel like a linter should not question the content of strings, although an argument can certainly be made otherwise. 

Also, I feel like `--dialect` should not have any effect on symbol/keyword parsing as that would make things more confusing.

jcburley:
> I think it'd be good to change Joker's symbol/keyword parsing to match Clojure behavior. As I understand this will not break any existing code as it allows a superset of symbols/keywords compared to what's currently allowed.

Submitted https://github.com/candid82/joker/issues/447 and have code nearly ready (though with Go's concept of whitespace at the moment). Executing that 1st seems wise at this point.

> I am OK with changing the option name to `:valid-ident` and I like the three options you proposed: `:any`, `:printable`, `:ascii`. The default could be `:ascii` as that's what most developers would want and this would allow to solve the original problem: warn on em dash.

Thinking more about this, I'd like to rename `:ascii` to `:alnum`, a term used in some regex providers to designate alphanumerics.

It would then be straightforward to explain, understand, and implement `:unicode-alnum`.

And as "printable" typically means to include spaces, which seems like something nobody would really want, I no longer think that's the right term for something that renders only as a visible image; `:graph`, `:grapheme`, or even `:graphemic` might be better. (This is getting really pedantic, and yes I did have to research this to dabble in the relevant distinctions. See for example [Unicode® Technical Standard #18](https://unicode.org/reports/tr18/).)

(I'm almost tempted to propose `:uses-toner` or `:displays-pixels` over any of the keywords proposed above!)

> What do you think we should do for those who do want larger Unicode set but still want to avoid em dashes and such? One solution could be adding the forth option: an arbitrary regex (for an individual symbol's character)... But that may be pretty annoying to deal with (you would have to know all Unicode classes that you want to allow etc). Another solution would be to add a separate option like `:ident-blacklist` that would list the characters that are explicitly forbidden.

Supporting just the proposed keywords and waiting for someone to present a decent use case for something different would give us an opportunity to support any combination of:

- one or a few more keywords with hardcoded meanings
- a Regex (instead of a keyword) that the user would have to code up as you point out
- an `:invalid-ident` keyword that would support a similar combination of keywords or a Regex and thus, if present, reject any characters (runes) allowed by the parser and (if present) the `:valid-ident` rule

> Not sure about `:valid-edn`. I'd rather not add anything without specific complaint / use case, and I am not 100% sure what problem that option would solve. The same goes for `:valid-string`. I feel like a linter should not question the content of strings, although an argument can certainly be made otherwise.

I don't grok EDN at this point, so am happy to punt that as you suggest.

Re `:valid-string`, at that point, anyone who wants only (their definition of) "vanilla" characters _inside_ the string probably feels the same way about other stuff, such as comments, conditionally excluded items, whatever. In that sense, strings aren't particularly "special", unless one considers `(symbol "arbitrary string with weird characters inside")`.

For such users, I think it makes more sense for them to use a distinct lint step, outside of Joker, to validate the source file as a whole, because information on Clojure/Joker tokens isn't needed. (That's what I would do if I wanted to impose such views on a dev org.)

> Also, I feel like `--dialect` should not have any effect on symbol/keyword parsing as that would make things more confusing.

That sounds reasonable. It's probably a lot less hassle to have Joker simply track Clojure, and let linting take care of those who want "stricter" Clojure, want to validate ClojureScript, etc., than to have Joker try to strike out on its own and define its own language beyond the necessities of being Go-based versus JVM-based.

candid82:
So after all the renaming you'd like to do, what would be the list of options and their semantics? 

>Supporting just the proposed keywords and waiting for someone to present a decent use case for something different would give us an opportunity to support any combination of...

Agreed. Let's postpone any work beyond a fixed set of options like `:any`, `:printables` and `:ascii` (or their equivalents).

Let's also punt on `:valid-edn` and `:valid-string` for now (I agree it probably makes more sense to use a separate checker for undesirable characters instead of introducing more options to a language specific linter).

jcburley:
> So after all the renaming you'd like to do, what would be the list of options and their semantics?

`:valid-ident` with a value of `:any`, `:alnum`, `:unicode-alnum` (or `:alnum-unicode`?), and hey how about `:visible` and `:unicode-visible` (or `:visible-unicode`).

`:alnum` would probably be a good default and would be the specific symbol characters allocated by Clojure (`*`, `+`, etc) plus `#"[a-zA-Z0-9]"`.

`:unicode-alnum` would extend `:alnum` with `unicode.IsLetter()` and `unicode.IsDigit()`.

`:visible` would extend `:alnum` with `IsGraphic(r) && !IsSpace(r) && r <= unicode.MaxASCII`, I think would be what I'm trying for (idea is to allow e.g. `$` but disallow space). Best described as "ASCII letters, marks, numbers, punctuation, symbols".

`:unicode-visible` would extend `:unicode-alnum` similarly, without the `unicode.MaxASCII` limitation.

If people want e.g. the e-acute`é` letter (`0xe9`) or the Cents symbol `¢`, (`0xa2`), but still no invisible or beyond-Latin1 codes, we could easily add `:latin1-alnum` and `:latin1-visible` to parallel the `:unicode-` ones.

Another popular choice might be to extend the `alnum` with just symbols (like `$`) but still no punctuation (like em dash). E.g. `:symbol` and `:unicode-symbol` would be straightforward.

This naming scheme would go on to support `:mark` and `:unicode-mark` as extending `:symbol` to include marks (still no punctuation), assuming nobody would want to allow marks but not symbols. I'd be quite surprised if anyone really needed these options, though.

A different approach would be to support a vector, so instead of `:unicode-alnum`, the value would be `:alnum` (meaning letters and digits with no other restrictions), and instead of `:alnum`, the value would be `[:ascii :alnum]` (in any order). That might be a tad harder to implement, but could be a little easier to document and comprehend, as well as extend elegantly. (The empty vector could denote whatever we pick as the default, but `:any` would still be supported.)

> > Supporting just the proposed keywords and waiting for someone to present a decent use case for something different would give us an opportunity to support any combination of...
> 
> Agreed. Let's postpone any work beyond a fixed set of options like `:any`, `:printables` and `:ascii` (or their equivalents).

Sounds good.

> Let's also punt on `:valid-edn` and `:valid-string` for now (I agree it probably makes more sense to use a separate checker for undesirable characters instead of introducing more options to a language specific linter).

Yup.

jcburley:
Thinking about the vector proposal, here's my reasoning (such as it is) for nomenclature.

In the context (such as ours) of specifying valid character sets for symbols, terms like "ASCII", "Latin1", and "Unicode" are "subtractive", in the sense that they specify limitations usually in relation to other selection(s) of character classes. ASCII means 7-bit, Latin1 means 8-bit, and Unicode means...well, anything in theory, but perhaps only actual Unicode character, not simply anything encodable via UTF-8 or whatever.

There aren't good (widely used) terms available that represent the "additive" forms of those terms, in the sense that nobody would expect `:ascii` to mean "any character in the 7-bit ASCII set, including control characters".

I.e. we want `:ascii`, `:latin1` (if we support that), and `:unicode` to limit the overall set of characters otherwise selected via terms pertaining more to character classes than to where they fall along the continuum of encoded values.

By contrast, `alnum` (or `alpha`, maybe `letter` to use modern terminology, plus `digit`), `symbol`, `mark`, `punctuation`, and `space` all specify character classes.

The single-keyword-as-value approach I first recommend, above, uses `alnum` and others to progressively extend valid symbol characters from the ones specifically allowed by Clojure, such as '+', `*`, `<`, and `?`, to include classes that "make sense" for such extension.

Everyone would want letters and digits in their symbol names, I assume, so that's the minimum restriction we'd support. (That is, nothing more restrictive than `:alnum`, as `:letter` or `:alpha` would seem rather useless, though easy enough to implement if I'm wrong about that.)

The next natural step in the progression would be to add symbols, hence `:symbol` would mean exactly that: letters, digits, and symbols.

In theory (I expect we'll never need this), marks would be next; then punctuation, which at that point I think basically becomes `:visible` in that it includes all "graphic" characters except spaces.

Inserting the `unicode-` prefix (or `-unicode` suffix), ditto with `latin1`, would allow these classes in the encoding ranges, with `ascii` (7-bit) being the default (or we could make `unicode` the default and support `ascii` explicitly).

With the vector approach, we'd probably want to split out these character classes for clarity, so `[:symbol]` would really mean just symbols. Seems strange except it's so much clearer to have people specify `[:letter :digit :symbol]` to get what the keyword-only denotes via `:symbol`.

But that'd mean the class-named keywords _add_ to the set of valid classes, while the encoding-named keywords _subtract_ from the resulting sets.

While strange, moving those encoding-named keywords into an `:invalid-ident` rule might be stranger, since (as explained above) we don't really have good concise terms for "characters beyond `MaxASCII`" or "characters beyond `MaxLatin1`.

So I'm in the fence regarding whether to support the vector approach versus just carving out specific, sometimes dual-named (encoding range plus set(s) of character class(es)), keywords.

jcburley:
Maybe the simplest thing for now is to support only "progressive", character-class, keywords for `:valid-ident`, and also support negative, encoding-range, keywords for an `:invalid-ident`. E.g.:

```clojure
{:rules {:valid-ident :symbol,
             :invalid-ident :non-ascii}}
```

While I'm not thrilled with the double-negative construction for `:invalid-ident`, it would at least be easy to explain. (I don't think any of these would be hard to implement, BTW.)

So for `:valid-ident`, character classes would be "progressive" per Joker's definition: possibly `letter` or `alpha`, then `:alnum`, then `:symbol`, then possibly `:mark`, then `:visible`, finally `:any`. (There are finer-grained distinctions available in there, but I don't know of any sensible progression they form.)

The "possibly"'s above denote things nobody probably wants, but I mention them for completeness.

If we later support a vector here, we'd probably have those same keywords lose their "progressive" qualities within the vector, and the implicit operation would be set union. E.g. `:symbol` would denote _only_ symbols, making `:valid-ident :symbol` mean the same thing as `:valid-ident [:letter :digit :symbol]`. Here, `:mark`, `:punctuation`, `:separator`, and so on would make more sense than in a progressive context.

`:invalid-ident` would support `:non-ascii`, possibly `:non-latin1`, `:non-unicode`, and `:any` (which would allow encodings beyond defined Unicode code points).

It could also support (ideally via a vector arg) the character-class keywords defined above, "notching" specific classes as _invalid_ despite being allowed by `:valid-ident`.

Either rules could someday support a Regex instead of a keyword/vector.

I think this approach provides the most-direct path to what people likely want now, in a way that should be easy enough to document and understand, and still provide a pathway to more-flexible capabilities should they prove worthy.

jcburley:
Closing in favor of https://github.com/candid82/joker/pull/450.

Issue #447, opened 2021-03-01
Valid Clojure Symbol/Keyword Names Rejected by Joker


jcburley:
Though it seems bizarre, Clojure accepts, as the initial and subsequent characters for a symbol/keyword name, any characters (encodable via UTF-8) that are not `java.base.java.lang.Character.isWhitespace()` and are not specifically designated to start another type of token or end a symbol/keyword name.

E.g. `(def ^@ "hi")`, where `^@` is how Emacs shows the NUL character, works; most "Control" characters in both the ASCII and extended-ASCII (Latin1) range work; and so on.

Joker currently allows only a limited set of characters in these circumstances: any specifically-designated character (`*`, `+`, and so on), Unicode Letter, or rune `r > unicode.Latin1` is accepted as an initial character, but not Control characters nor Latin1 characters. Especially in light of the Latin1 test (currently directly encoded as the manifest constant `255`), this appears to be somewhat arbitrary; but for characters in the ASCII and Latin1 ranges, it's reasonable enough.

Further, Joker's use of Go's `unicode.IsSpace()`, to detect whitespace, differs subtly from that `isWhitespace()` function used by Clojure, as referred to above, in that non-breaking whitespace is considered whitespace by the former but not by the latter. Whether Java's definition is correct is a subject of some dispute (I won't bother putting links here); but strict conformance to this would require identifying and excluding all non-breaking space in Unicode, something Unicode itself doesn't seem to provide directly (at least not according to the source code of Go's `unicode` package that I can see).

As discussed [here](https://github.com/candid82/joker/pull/445#issuecomment-787617243), it seems best to have Joker match (very closely if not "perfectly") Clojure's behavior, and leave it to new linter options (also discussed in that comment and PR generally) to narrow down the permissible character sets to various reasonable subsets.

I think fixing this Issue first is the best way to proceed; from there, clarity should only increase regarding linter options in that other PR (and the [discussion](https://github.com/candid82/joker/discussions/444) that kicked this all off in the first place.

PR #448, opened 2021-03-01
Allow Symbol/Keyword Lexemes a la Clojure

jcburley:
Fixes https://github.com/candid82/joker/issues/447, except for the non-breaking-space exceptions described therein.

### DIFF
--- a/core/read.go
+++ b/core/read.go
@@ -408,6 +408,8 @@ func readNumber(reader *Reader) Object {
 func isSymbolRune(r rune) bool {
 	switch r {
 	case '"', ';', '@', '^', '`', '~', '(', ')', '[', ']', '{', '}', '\\', ',', ' ', '\t', '\n', '\r', EOF:
+		// Whitespace listed above (' ', '\t', '\n', '\r') purely for speed of common cases
+
 		return false
 	}
 	return !isJavaSpace(r)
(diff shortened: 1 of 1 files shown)


jcburley:
Should I add a comment to the `case` statement in `isSymbolRune` explaining that a few whitespace characters (as well as comma), which could be left to the fallthrough case, are listed in a quest for a bit of extra speed, as they're so common in source code?

jcburley:
Short of writing a Clojure (or Java) program enumerating all of the Unicode points along with their corresponding Character.isWhitespace() predicates, and comparing that to what the new `isJavaSpace()` function does, this is about as close as we need to get for now.

candid82:
> Should I add a comment to the `case` statement in `isSymbolRune` explaining that a few whitespace characters (as well as comma), which could be left to the fallthrough case, are listed in a quest for a bit of extra speed, as they're so common in source code?

Yes, I think a comment would not hurt. Then it will be good to go. Thank you for being so thorough in handling this!

jcburley:
Okay, that should do it for now. I might return to this (and write Clojure/Java and Joker/Go programs to each summarize which runs are whitespace) down the road, but I think it's more urgent to get in the linter improvements next.

candid82:
Looks great, thank you!

Issue #252, opened 2019-08-23
Guide for adding new namespaces?


rcarmo:
Hello there - I'm trying to create my own namespace to extend `joker`, and I can't quite figure out how to go about it.

I've cloned the repo, added a new folder inside `std` and a matching `.joke` file, wrote `a_` and `_native` stubs for what I want to invoke and modified `generate-std.joke` to add the new namespace, but it doesn't get compiled in.

Is there a documented way to add new functionality to Joker?

Thanks!

jcburley:
See https://github.com/candid82/joker/blob/master/core/data/README.md, which we should modify based on your experiences getting it to work.

E.g. you shouldn't have to write an `a_*.joke` file yourself -- it's autogenerated by `generated-std.joke`.

jcburley:
BTW, what does your new namespace do?

rcarmo:
Thanks! I'll look into it. I'm trying to write a wrapper for [sqlite-lite](http://github.com/bvinc/go-sqlite-lite), since #245 was closed, I would find it extremely useful and it seems straightforward enough when compared to other `sqlite` libraries.

I'll report back when I have reproducible steps or feedback for that README file...

pmonks:
sqlite support would be great!  I'm a little nervous about introducing a requirement on `cgo` however, since my (limited!) understanding is that that prevents cross-compilation (something I'm doing with `joker`).

rcarmo:
Well, I’m using a library that does work on the Raspberry Pi, and developing on one, so that’s part of my goals.

pmonks:
My understanding is that *cross compilation* is substantially more difficult once `cgo` is in the mix, even when the C libraries that are being interfaced with are available on the target platform.

Happy to be proven wrong, but just want to make it clear that there's value (to me, at least) in preserving the ease with which `joker` can be cross-compiled.

rcarmo:
Well, I have a fair bit of experience with cross-compilation (not Go, plain C/C++) and have of late simplified matters a great deal for my container workflows: https://github.com/rcarmo/azure-pipelines-multiarch-docker/blob/master/README.md

I’m sure joker can ultimately built by native toolchains in an emulated CPU if we have any trouble with cgo...

candid82:
I think this issue is now addressed by https://github.com/candid82/joker/blob/master/DEVELOPER.md (thanks @jcburley!)

rbucker:
[...]

Issue #245, opened 2019-08-06
SQLite support?


rcarmo:
I realize this would mean a (beefy) extra dependency like [go-sqlite3](https://github.com/mattn/go-sqlite3) (which requires `cgo` and more testing), but accessing SQLite databases would be an awesome addition to Joker's scripting capabilities.

candid82:
I understand the appeal of having built-in SQL engine for some use cases, but I think it would bring the disproportional amount of additional complexity. Perhaps exploring CGO (which I don't know much about) and potential uses of it in Joker should be the first step here... If anyone wants to experiment with the idea, I'd be curious to see what it could look like, but I probably won't work on things like this any time soon.

pmonks:
My (extremely limited!) understanding is that the use of `cgo` would limit cross-compilation, which would be a real shame.  I'm using `joker` on an old Raspberry Pi (cross compiled from macOS), and it works very well.  I'd be sad to lose that.

PR #449, opened 2021-03-01
RFC: Markdown Rendering Support

rcarmo:
Hello there, 

I have been using `joker` for a while now and have found it extremely useful for automating a number of chores, but needed it to be able to render markdown, so I did some research into the _status quo_ of Go parsers and found  [`goldmark`](https://github.com/yuin/goldmark) to both have everything that was required and also a very nicely laid out approach.

I've tried adding other things to `joker` in the past (most notably SQLite - see #245 and #252), and this is comparatively much less complex, but perhaps more generally useful (and yes, I'll try having a go at SQLite again sometime soon...)

I've added docs as well (such as they are), but am uncertain of how to go about doing test coverage - so feel free to comment.


candid82:
@rcarmo  This looks interesting. I'll look at the specifics of the implementation, but my main concern is how "heavy" that `goldmark` dependency is. How much does it add to the size of the binary?

rcarmo:
Around 3MB, I would say. I would need to rebuild mine on Linux to get an accurate sizing. It did not seem like a big consideration since binary sizes vary a bit across architectures (and can be dealt with by using upx and the like).

borkdude:
Another option would be to implement something like babashka pods which make it possible to create optional libraries from other standalone executables. This is how babashka keeps the binary size down while giving access to a whole range of libraries, even go-sqlite3.

https://github.com/babashka/pod-registry

candid82:
@borkdude Yeah, I think pods are a very cool idea. I like their flexibility (you can write them in any language). However, something doesn't sit right with me about the fact that what looks like a library call actually does IPC under the hood. There are obvious performance implication (I know, performance is not very important for Joker or babashka, but still), but I guess I am more concerned with the hidden complexity here. It's more like a gut feeling than a real argument though. Perhaps I need to explore this idea more. 
I wish Go officially supported dynamic library loading on all platforms. Although Go's plugins are now supported on Linux and MacOS, there is still no Windows support. However, according to github download counts, Windows is only 2-3% of downloads, so perhaps supporting (dynamically loaded) Go libraries would be another viable alternative...

Getting back to the actual RFC, I think the question is whether or not markdown rendering belongs in the built-in standard library. For "regular" languages the answer would probably be no. However, Joker strives to be "scripting friendly", which, among other things, means having readily available tools to support common "scripting" tasks. Now, the definition of "scripting" is very vague and I am not sure if "markdown rendering" can be considered a common enough task, but I myself have had a few use cases in the past for markdown rendering, so it's definitely not rare... 
So I am leaning toward including it in the standard library.
@jcburley do you have any opinion here or any thoughts on pods or dynamic library support in general?

jcburley:
I like the idea of Joker having a plethora of commonly-used packages for scripting, but am uneasy about not only the impact on the executable size, but also the longer-term issues around dependencies on 3rd-party libraries that might not be kept up over time.

Joker already has seemingly-necessary dependencies on 3p packages for things like reading lines, yet has already switched those "midstream" with subsequent changes in behavior that had to be managed. It has since added `bbolt`, which (though 3p) is strikes me as an excellent example of making Joker more useful for scripting without much overhead.

The Go std library itself has the "don't break anything" rule (in v1 anyway), besides obviously being well-maintained by dint of having funding, etc. The further away Joker's dependencies are from that nature, the more risk I think we take on.

Note that, last I checked, the Go people hadn't fixed whatever causes MacOS plugin support to break things. (I found and submitted the bug myself a year or so ago). So I'm not sure we want to rely on plugins just yet, although they look quite promising to me, and I suspect would be fairly straightforward to "wrap" dynamically (in Joker) _a la_ how my `gostd` fork statically wraps packages.

So my `gostd` fork, which is designed to (at least somewhat) "solve" the 3p problem by automating the addition of wrappers around arbitrary packages (though currently only the Go std is supported), actually excludes the 'plugin' package when built on MacOS. (Otherwise the resulting executable can randomly hang.)

Re IPC/RPC underlying function calls, it's certainly a controversial topic in general, but something like Joker would IMO truly benefit from having such a mechanism, if sufficiently generalized, just as it would from plugins. Joker seems well-suited to, and appropriate for, orchestrating various services across processes and machines; and Clojure/Joker support for decent APIs (including things like futures, lazy evaluation, etc) surely helps.

I haven't more than glanced at pods, though, so I don't know how complicated they are to implement; but I've been using (and to some extent implementing) IPC/RPC-like mechanisms for a long time now, so to me the only concern here is the specifics as they relate to Joker, being written in Go and all, not the idea itself.

borkdude:
FWIW, this is how you can get markdown rendering today in babashka:

``` clojure
(require '[babashka.pods :as pods])

(pods/load-pod 'retrogradeorbit/bootleg "0.1.9")

(require '[pod.retrogradeorbit.bootleg.markdown :as md])

(def md-text
"# Hello

- This is a list
- Another item"
  )

(md/markdown md-text :data :html)
```

```
"<h1>Hello</h1><ul><li>This is a list</li><li>Another item</li></ul>"
```

The [pod protocol](https://github.com/babashka/pods) is open for anyone to implement and will give you access to a collection of already [existing pods](https://github.com/babashka/pod-registry).

It would be nice if we can share the protocol and joker will not re-invent its own, so "pods" can be re-used for both scripting environments.

borkdude:
The same can be accomplished in joker today by just shelling out to the [bootleg](https://github.com/retrogradeorbit/bootleg) program:

```
(require '[joker.os :as os])

(def md-text
  "# Hello

- This is a list
- Another item"
  )

(def convert-markdown (list 'markdown md-text :data :Html))

(prn (:out (os/exec "bootleg" {:args ["-e" (pr-str convert-markdown)]})))
```

```
"<h1>Hello</h1><ul><li>This is a list</li><li>Another item</li></ul>"
```

After all, the pod RPC is basically a glorified way of shelling out.

borkdude:
Some proof that pod clients (where the pod server = the programming offering a library over RPC) can be implemented in other languages:

- Common Lisp: https://github.com/rorokimdim/stash/blob/master/scripting-examples/common-lisp/stash.lisp
- Python: https://github.com/rorokimdim/stash/blob/master/scripting-examples/python/stash.py

The pod registry is enough proof that pods (the libs) can implemented in any language and has examples of golang as well.

rcarmo:
Hi guys, It seems I've sparked an animated discussion. 

Here's my take on this - I need to do a fair amount of scripted content generation and data munging (the bigger picture is getting JSON and XML data in, snipping out bits of it, then merging that with various source docs and spitting out results). And having a single binary to do those and script it sanely is appealing for various reasons, but the parsing and rendering needs to be fast, so I started with the output phase (which is easier and more generically useful, since I'm pretty sure everyone needs to generate content anyway).

I'm also looking at XPath and JSONPath functionality and trying to find libraries that would be a good fit (both maintainability-wise and quality-wise) to build out the rest (ideally parsing the data into nice sexps we can play with).

I do understand the concerns around binary size (which is another reason why I started with a common-sense, generally useful thing like Markdown) and modularity (which I suppose is a broader discussion), just explaining my viewpoint.

borkdude:
@rcarmo Why must all the functionality live inside one binary? I compare joker and babashka to replacements of bash scripts. In bash scripts it's very natural to leverage other binaries to accomplish a task. Using `clojure.java.shell/sh` or `java.os/exec` you can have a wider reach of tools without building a giant kitchen sink monolithic tool. And this is still fast.

rcarmo:
Because I want to deploy _one thing_ to get work done, ideally. Today I can do that with a `.jar` file, but need the entire JVM there as well. And I don't always control where my stuff gets used.

borkdude:
Babashka pods download automatically when you load them and aren't there yet. Does that fit with your "deploying of one thing"? 

If not, and if joker doesn't has or accept all the things you need, I guess you could always roll your own minimal Go CLI.

rcarmo:
Not everything has Internet access.

jcburley:
Among Joker's big draws (or at least somewhat-unique characteristics) in the Clojureverse are its single, fairly compact executable; reasonably quick startup; and the lack of reliance on 3rd-party libraries that are downloaded at runtime.

These characteristics help avoid dependency/update problems experienced in the JavaScript world, among others; they also keep startup time quick due to lack of reliance on dynamic downloading, use of plugins/pods, etc to obtain whatever one might consider "basic" functionality for scripting.

I'm not sure what appeal Joker would have, beyond perhaps being a fast linter, to those who don't care about those characteristics (though am curious to find out!). Wouldn't Clojure itself meet their needs just fine, without any of the hassles of running in a "not-quite-fully-Clojure" environment?

And fast startup (if not compact executable), at least, is delivered by Babashka, which (last I checked) is just about as quick as Joker, though slightly more resource-intensive, probably due to being built on a JIT (GraalVM) that speeds things up over time but adds some overhead.

So while Joker's utility would be _greatly_ expanded by natively supporting pods, plugins, and other-such mechanisms (just as it supports shelling out), I think it makes good sense to preserve its unique characteristics while adding capabilities _that make sense for a general-purpose scripting tool_.

I.e. for anything (available as a Go package at least) that is reasonably useful in a scripting tool, regardless of specific scripts being run, one _should_ be able to build, if not simply download, a Joker executable that includes it "natively" (without relying on dynamic mechanisms).

Even then, it's not clear what should be in "official Joker" versus what people should be expected to "build in" (by whatever mechanism) on their own, should they decide the official version doesn't meet their needs.

In fact, the executable-size (and, perhaps to a lesser extent, startup-speed) issue could be flipped on its head: official Joker _could_ be generous when it comes to adding reasonably popular, well-maintained namespaces (libraries) that make sense for a scripting language and "want" to be quickly and seamlessly accessible (no dynamic loading, no network needed); and yet support easy re-building _without_ nearly any of those namespaces that a given shop doesn't want, in order to tighten up Joker's executable size, startup speed, and possibly even security profile. (Right now I think it's fairly easy, though not fully automated, to remove a namespace like `joker.bbolt` and rebuild Joker so the executable includes none of that code. Does it need to be better documented? Fully automated?)

At some point, assuming Joker continues to grow in utility and audience, it could make sense to spin up an official repository to contain user-contributed libraries/namespaces like other languages have (repos such as npm, elpa, and so on). Perhaps it'd need to formally distinguish between build-time (drop-in) libraries, pods, plugins, and so on.

Towards that end, it'd be helpful if Joker was as accommodating as possible for such repos. Work done awhile back on `*ns-sources*` (see [`LIBRARIES.md`](https://github.com/candid82/joker/blob/master/LIBRARIES.md)) is among things already accomplished; I've suggested (and given some thought to) uniting the `std` and `core/data` namespace-incorporation mechanisms Joker currently supports to provide a single, straightforward mechanism into which arbitrary namespaces could be downloaded and dropped; perhaps more or different things could be done.

(And I think that's all orthogonal to my `gostd` fork, which automatically wraps, with Joker "goo", Go packages for static inclusion into a Joker executable. It is more at the level of using, maybe someday first dynamically downloading, Go packages, rather than Joker libraries.)

There's lots to think about and ultimately accomplish here. But it's great to have something as malleable a Joker is (built on Go, which for me strikes a reasonable balance among various concerns), implementing a Lisp-like language as "pure" and well-thought-out as is Clojure, oriented for shell-like scripting. I've been dreaming about (and playing around with designing) such a thing for a couple of decades, taking all the above concerns into account, and was thrilled when I discovered Joker, just in time to avoid trying to implement Clojure in Go!

borkdude:
Just for the record, babashka is not a JIT interpreter/compiler: I think the model is pretty similar to joker, with the difference that it's more JVM-focused and written in Clojure itself.

jcburley:
> Just for the record, babashka is not a JIT interpreter/compiler: I think the model is pretty similar to joker, with the difference that it's more JVM-focused and written in Clojure itself. Compiled to native using GraalVM native-image.

Oh, interesting! I thought Babashka incorporated GraalVM and the result was pretty much just a leaner, meaner JVT (complete with JIT).

Re-skimming a couple of pertinent web pages, now it seems more like Babashka uses GraalVM as a compiler for itself, but doesn't incorporate it to provide JIT to Clojure code at runtime.

It's so similar to Joker, from what little testing I've done, as far as startup time and such, that the only differences I've noticed is that it seems to need other artifacts to be installed somewhere (I can't recall the details, but it looked like just a single executable wasn't enough) and that it provides fairly complete access to the JVM (and thus Clojure) ecosystem, which Joker (currently) does not.

What advantages does Joker have (in theory if not currently) over Babashka?

borkdude:
@jcburley 

> that it seems to need other artifacts to be installed

Not sure what you mean here. Maybe compared to golang: golang executables are fully standalone, GraalVM binaries can be dynamically linked (needs libc++ installed, but this is available on most systems already usually) or statically compiled (for linux only). Most users can just download the dynamically linked one and run it without any problems, except on Alpine, there you need the static one.

>  provides fairly complete access to the JVM 

I guess that is the main difference between joker and babashka. Babashka allows interop to (built-in) classes like `java.io.File`, etc and can handle a list of libraries from clojars. Some of them are listed here: https://github.com/babashka/babashka/blob/master/doc/projects.md. Babashka cannot run bytecode, only source. This is where pods come in: you can write a pod in any language and then interact with it from babashka.

Babashka's interpreter is split out into a library: https://github.com/borkdude/sci
This allows you to write a GraalVM scripting binary like babashka but with your own selection of libraries. Bootleg (already mentioned before here in relation to Markdown) is an example of this.

I am not familiar enough with Golang in general (I only dabbled with it to create a go-sqlite3 pod to use from babashka), so I can't really say what benefit the one ecosystem offers over the other, but it's great to have Clojure scripting for both!

To not hijack this thread too much with off topic discussion which is not related to this RFC, you can find me on Slack, etc if you want to know more.

I suggest we get back to the original topic.

borkdude:
One more idea:

Babashka also has feature flags, so if you decide to compile babashka yourself, you can get extra features (or less features) based on these flags.

https://github.com/babashka/babashka/blob/master/doc/build.md#feature-flags

E.g. `BABASHKA_FEATURE_POSTGRESQL` is disabled by default, but when enabled, you can use `next.jdbc` to interact with postgres.

That might also an approach which joker could adopt, e.g. `JOKER_FEATURE_MARKDOWN`.

jcburley:
> @jcburley
> 
> > that it seems to need other artifacts to be installed
> 
> Not sure what you mean here. Maybe compared to golang: golang executables are fully standalone, GraalVM binaries can be dynamically linked (needs libc++ installed, but this is available on most systems already usually) or statically compiled (for linux only). Most users can just download the dynamically linked one and run it without any problems, except on e.g. Alpine Docker containers, there you need the static one.

Ah, sorry, not sure how I formed my impression of that; thought I'd seen some other artifacts installed on my laptop (or desktop) after installing bb awhile back, but guess not?

> > provides fairly complete access to the JVM
> 
> I guess that is the main difference between joker and babashka. Babashka allows interop to (built-in) classes like `java.io.File`, etc and can handle a list of libraries from clojars. Some of them are listed here: https://github.com/babashka/babashka/blob/master/doc/projects.md. Babashka cannot run bytecode, only source. This is where pods come in: you can write a pod in any language and then interact with it from babashka.

They (Joker and Babashka) really are more alike than I'd thought!

> Babashka's interpreter is split out into a library: https://github.com/borkdude/sci
> This allows you to write a GraalVM scripting binary like babashka but with your own selection of libraries. Bootleg (already mentioned before here in relation to Markdown) is an example of this.

I should play around with Babashka more then.

> I am not familiar enough with Golang in general (I only dabbled with it to create a go-sqlite3 pod to use from babashka), so I can't really say what benefit one ecosystem offers over the other, but it's great to have Clojure scripting for both!

Indeed. While I think of the JVM ecosystem as very mature, having a Clojure-like scripting engine for the Golang ecosystem seems like a win.

> To not hijack this thread too much with off topic discussion which is not related to this RFC, you can find me on Slack, etc if you want to know more.

Yup. Are you thinking of a specific channel? Of course I'm on `#joker` as well as `#clojure`, `#clojurescript`, `#lumo`, `#closh`, `#graalvm`, and of course `#off-topic`.

> I suggest we get back to the original topic.

But I learn so much from off-topic discussions!! ;)

jcburley:
> One more idea:
> 
> Babashka also has feature flags, so if you decide to compile babashka yourself, you can get extra features (or less features) based on these flags.
> 
> https://github.com/babashka/babashka/blob/master/doc/build.md#feature-flags
> 
> E.g. `BABASHKA_FEATURE_POSTGRESQL` is disabled by default, but when enabled, you can use `next.jdbc` to interact with postgres.
> 
> That might also an approach which joker could adopt, e.g. `JOKER_FEATURE_MARKDOWN`.

Nice!

borkdude:
@jcburley 

> Yup. Are you thinking of a specific channel? Of course I'm on #joker as well as #clojure, #clojurescript, #lumo, #closh, #graalvm, and of course #off-topic.

There is also #babashka :-)

borkdude:
@jcburley @candid82 Also keep an eye on https://github.com/nooga/let-go. 

> This is supposed to be a compiler and bytecode VM for a language resembling Clojure as close as possible.

candid82:
Alright, I am going to merge this PR because I generally agree with what James wrote here:

> In fact, the executable-size (and, perhaps to a lesser extent, startup-speed) issue could be flipped on its head: official Joker could be generous when it comes to adding reasonably popular, well-maintained namespaces (libraries) that make sense for a scripting language and "want" to be quickly and seamlessly accessible (no dynamic loading, no network needed); and yet support easy re-building without nearly any of those namespaces that a given shop doesn't want, in order to tighten up Joker's executable size, startup speed, and possibly even security profile. 

The goal is to make Joker as useful as possible while keeping it (and its usage) as simple as possible. This is, of course, a very vague goal and everything essentially boils down to personal preference :)

The concern about yet another 3p dependency is a valid one. However, `goldmark` seems to be "self contained" (doesn't have other deps), which makes it somewhat less of a concern. 

I do think that it'd be good to allow "advanced" users to easily build custom Joker binary with some standard namespaces excluded, perhaps using a mechanism similar to babashka's feature flag. But no one has actually asked for this yet, so maybe it's not that important (yet).

Dynamic library loading is a separate big topic. Looks like plugins are not ready for prime time yet. It'd be interesting to explore pods support more, even though I am not convinced that it would be widely used by Joker's users even if implemented. 

Thanks @jcburley and @borkdude for your thoughts! Thanks @rcarmo for the contribution!

rcarmo:
No, thank you. I am now looking for a good XPath (and JSONPath) library for the rest of my pipeline. Will come back with those once I find suitable ones.

PR #450, opened 2021-03-01
Lint Identifiers

jcburley:
This documents my latest thinking on how best to address https://github.com/candid82/joker/discussions/444 in light of recent discoveries and changes to Joker.

The discussion should probably move over to here from https://github.com/candid82/joker/pull/445, which I'll Close soon, as it's fairly obsolete at this point.


README.md
```diff
@@ -233,29 +233,40 @@ When linting directories Joker lints all files with the extension corresponding
...
-| `fn-with-empty-body`   | warn on fn form with empty body                       | `true`        |
+| Rule                   | Description                                             | Default value |
+|------------------------|---------------------------------------------------------|---------------|
+| `if-without-else`      | warn on `if` without the `else` branch                  | `false`       |
+| `no-forms-threading`   | warn on threading macros with no forms, i.e. `(-> a)`   | `true`        |
+| `unused-as`            | warn on unused `:as` binding                            | `true`        |
+| `unused-keys`          | warn on unused `:keys`, `:strs`, and `:syms` bindings   | `true`        |
+| `unused-fn-parameters` | warn on unused fn parameters                            | `false`       |
+| `fn-with-empty-body`   | warn on fn form with empty body                         | `true`        |
+| `valid-ident`          | warn on symbols/keywords with non-special characters    | `:alnum`      |
```
candid82, line 259: I like the semantics of these proposed options, specifically the separation between character classes and encoding ranges (and thank you for writing down your thought process in https://github.com/candid82/joker/pull/445 that led you to this proposal, it was really helpful). 
I agree that double negative in `invalid-ident` looks unnatural. Also, `valid-ident` and `invalid-ident` are not symmetrical (despite what their names would suggest), as the former talks about character classes and the latter talks about encoding ranges. So, what if we make the names more closely follow the semantics:
```
{:valid-ident {:character-class :alnum
               :encoding-range :ascii}}
```
This looks more natural to me. What do you think?

core/read.go
```diff
@@ -475,6 +476,124 @@ func readSymbol(reader *Reader, first rune) Object {
...
+			i = iValid
+			explain = explainValid
+		} else if iValid > iInvalid {
+			i = iInvalid
+			explain = explainInvalid
+		} else { // iValid == iInvalid
+			i = iValid
+			explain = explainValid + "; " + explainInvalid
+		}
+	}
```
candid82, line 554: The above looks more complicated than it could be... Would not it be easier to have predicate functions that check the validity of an individual rune (that it belongs to the selected character class(es) and to the selected encoding range), and then do one loop over the identifier string and report the first invalid rune? Something like 
```
func warnInvalidString(s *string) {
	for _, r := range s {
		if !isIdentSpecificRune(r) && !(isValidIdentCharacterClass(r) && isValidIdentEncodingRange(r)) {
			// print warning
			return
		}
	}
}
```

README.md
jcburley: Ooh, I like that a lot better, it's so clear and concise! Yet it still allows for flexibility down the road.

core/read.go
jcburley: Ignore that code you quoted; it was indeed unnecessarily complicated, mainly an attempt to just make nice diagnostics where they weren't really needed.

I'll update the PR with your proposal in the README.md, and (when I get time; very busy these next several days) provide a cleaner, and complete, implementation that I hope won't be too slow.

jcburley:
Updated the PR. I think `:core` is a better keyword than `:alnum`. What do you think?

candid82:
Yes, I like `:core` and `:character-set`. But now I am having a second thought on whether `valid-ident` should be inside `rules`. All other rules are boolean flags... So perhaps it's better to have `valid-ident` as a root level key rather than inside `rules`? It will probably make it easier to document it without having to adhere to the pattern for `rules` ("warn on <something>").

jcburley:
Great idea, this is so much cleaner! Let me know how it looks and feels to you, based on my revisions to README.md herein.

jcburley:
There is one more thing I've been wondering about: if possible, would it make sense to limit this checking to only contexts in which a _new_ symbol or keyword is being defined? (I guess one doesn't really reference versus define keywords, so this really applies just to symbols.)

One might wish to detect attempts to define "weird" symbol names in one's own namespaces while allowing references to such weird names in others'.

It would be amazing if Joker reached the sort of audience(s) that made such an approach useful in the real world, but I thought I'd throw that idea out....

candid82:
The docs look great! As for contextual checking... this is an interesting idea but is probably an overkill. We can always do it later though if there is a real need. That, however, gave me an optimization idea (not that it will make a huge difference): to avoid checking every identifier (as there may be lots of duplicates) we could keep the ones we already checked in a map (with keys of type `*string`, which should be pretty efficient). Again, not sure if it will make any difference, but maybe worth trying.

jcburley:
> The docs look great! 

Okay, will proceed with implementation.

>As for contextual checking... this is an interesting idea but is probably an overkill. We can always do it later though if there is a real need.

Makes sense to me!

>That, however, gave me an optimization idea (not that it will make a huge difference): to avoid checking every identifier (as there may be lots of duplicates) we could keep the ones we already checked in a map (with keys of type `*string`, which should be pretty efficient). Again, not sure if it will make any difference, but maybe worth trying.

Intriguing, but that'd mean only the first encounter (which might not necessarily be the first definition) of a symbol would yield warnings.

If there was to be only one warning per symbol name in a linter run, ideally it'd be where it's defined.

But I think many (most?) users would expect warnings for each occurrence in their source code.

BTW, do you think there should be only one warning per identifier (about the first problematic encounter encountered), or should the scanning continue and subsequent warnings be issued? If verbosity is a concern,  a single warning showing the set of problematic characters (and pointing to the first one) wouldn't be hard to do.

jcburley:
This is now implemented such that each problematic character is warned about (the diagnostic nicely explains what the violation(s) is/are).

candid82:
> Intriguing, but that'd mean only the first encounter (which might not necessarily be the first definition) of a symbol would yield warnings.

Not necessarily, you can still warn on every occurrence. You just don't have to re-check strings that are already cached in a map (with all the details about what's wrong with the string, if anything).

Again, totally optional optimization (if an optimization at all).

It's getting late, I'll give a final look tomorrow before merging.

jcburley:
> > Intriguing, but that'd mean only the first encounter (which might not necessarily be the first definition) of a symbol would yield warnings.
> 
> Not necessarily, you can still warn on every occurrence. You just don't have to re-check strings that are already cached in a map (with all the details about what's wrong with the string, if anything).

Ah, yes, that'd be transparent to the user (and easy enough to do). 

> Again, totally optional optimization (if an optimization at all).

`time ./linter-tests.sh` doesn't show me any significant difference in performance between this PR and `master`. I guess Joker does so much other work when linting that this new stuff ends up in the noise?

> It's getting late, I'll give a final look tomorrow before merging.

Great, thanks!

core/read.go
```diff
@@ -475,6 +475,144 @@ func readSymbol(reader *Reader, first rune) Object {
...
+		if !isCoreIdentRune(r) && (!identValidationSetFn(r) || !identValidationRangeFn(r)) {
+			var explain string
+			if identValidationSetFn(r) {
+				explain = identValidationRangeWhy
+			} else if identValidationRangeFn(r) {
+				explain = identValidationSetWhy
+			} else {
+				explain = identValidationSetWhy + "; " + identValidationRangeWhy
+			}
+			runeValue, _ := utf8.DecodeRuneInString((*s)[i:])
```
candid82, line 555: Would not `runeValue` be the same as `r`?

candid82:
Thanks!

core/read.go
jcburley: > Would not `runeValue` be the same as `r`?

Ah, I'm sure it would now that the code works differently than the earlier version!

I'll do up a PR soon, after I take care of other things; or you can try it yourself.

(Which reminds me, I haven't gotten around to adding tests for this new linting capability yet; I just run my own hand-rolled ones....)

candid82: No worries, I'll make the change and add tests.

candid82: Made a few tweaks here: https://github.com/candid82/joker/commit/f3dd4c864978ef55330a43c31db279f1faa043f7

PR #451, opened 2021-03-01
Fix these independently of linter PR(s)

jcburley:


candid82:
I did not use curly braces to emphasize that those keys are not necessarily the only keys in the config map, but I guess it's less confusing to have curly braces. Thanks!

jcburley:
> I did not use curly braces to emphasize that those keys are not necessarily the only keys in the config map, but I guess it's less confusing to have curly braces. Thanks!

No problem. I saw it mainly as a consistency issue; there were several other examples of `.joker` snippets above those two, all showing the entire map (i.e. the entire file). Now they all show an entire map.

Issue #452, opened 2021-03-02
Joker Crashes Linting a ClojureScript Source File


jcburley:
E.g.:

```
$ joker --lint /Users/craig/github/clojure/clojurescript/src/test/clojure/cljs/analyzer_tests.clj
/Users/craig/github/clojure/clojurescript/src/test/clojure/cljs/analyzer_tests.clj:48:42: Parse warning: unused binding: extra
/Users/craig/github/clojure/clojurescript/src/test/clojure/cljs/analyzer_tests.clj:56:32: Parse warning: unused binding: name
/Users/craig/github/clojure/clojurescript/src/test/clojure/cljs/analyzer_tests.clj:59:35: Parse warning: unused binding: name
/Users/craig/github/clojure/clojurescript/src/test/clojure/cljs/analyzer_tests.clj:190:7: Parse warning: fn form with empty body
/Users/craig/github/clojure/clojurescript/src/test/clojure/cljs/analyzer_tests.clj:445:11: Parse warning: unused binding: test-env
/Users/craig/github/clojure/clojurescript/src/test/clojure/cljs/analyzer_tests.clj:608:30: Parse error: Unable to resolve symbol: a.b.c
/Users/craig/github/clojure/clojurescript/src/test/clojure/cljs/analyzer_tests.clj:608:25: Parse warning: unused binding: a
/Users/craig/github/clojure/clojurescript/src/test/clojure/cljs/analyzer_tests.clj:620:25: Parse warning: do form with empty body
/Users/craig/github/clojure/clojurescript/src/test/clojure/cljs/analyzer_tests.clj:621:25: Parse warning: redundant do form
/Users/craig/github/clojure/clojurescript/src/test/clojure/cljs/analyzer_tests.clj:625:19: Parse warning: do form with empty body
/Users/craig/github/clojure/clojurescript/src/test/clojure/cljs/analyzer_tests.clj:626:19: Parse warning: redundant do form
/Users/craig/github/clojure/clojurescript/src/test/clojure/cljs/analyzer_tests.clj:627:19: Parse warning: redundant do form
/Users/craig/github/clojure/clojurescript/src/test/clojure/cljs/analyzer_tests.clj:630:19: Parse warning: let form with empty bindings vector
/Users/craig/github/clojure/clojurescript/src/test/clojure/cljs/analyzer_tests.clj:630:19: Parse warning: let form with empty body
[...]

PR #453, opened 2021-03-02
Stop crashing for now

jcburley:
Fixes https://github.com/candid82/joker/issues/452.

But I suspect this merely hides a bug somewhere...?


candid82:
Thanks! I'll take a deeper look for the root cause later.

PR #457, opened 2021-03-30
Better native conversion for (format ...)

jcburley:
This helps `(format ...)` produce useful results for format specifiers other than `%s`, such as `%x`, `%f`, and `%g`, when the operands don't fit into Native Joker types but can be accurately represented as native Go types.

The toNative() function is renamed so it's exported and thus usable by other libraries in the future.


candid82:
Thanks!

PR #458, opened 2021-04-01
Use math/big Package Properly

jcburley:
The docs for big.Float, big.Int, and big.Rat specify that objects of those types shouldn't be copied (except via the appropriate `.Set()` method), but rather that pointers to them be used:

> shallow copies of <type>s are not supported and may lead to errors

It seems best (and fairly straightforward) to conform to those requirements.


candid82:
Thanks!

PR #459, opened 2021-04-01
Improve Diagnostics For Unexpected Crashes When Reading

jcburley:
While debugging my own changes to Joker, I kept running into this,
which I found rather inscrutable:

```
user=> 1234N
runtime error: invalid memory address or nil pointer dereference
user=>
```

Without the stack trace, I was left nearly bereft of useful info.

This adds the stack trace for when the error isn't an expected one
from other `panic`s in the code. E.g.:

```
user=> 1234N
<repl>:1:5: Read error: Unexpected panic "runtime error: invalid memory address or nil pointer dereference" at:
goroutine 1 [running]:
runtime/debug.Stack(0x166e520, 0x1668cc0, 0x1b7c1a0)
	/usr/local/go/src/runtime/debug/stack.go:24 +0x9f
github.com/candid82/joker/core.TryRead.func1(0xc0004bf988, 0xc0000c0000)
	/Users/craig/go/src/github.com/candid82/joker/core/read.go:1499 +0xbe
panic(0x1668cc0, 0x1b7c1a0)
	/usr/local/go/src/runtime/panic.go:965 +0x1b9
math/big.(*Int).scan(0x0, 0x1887c40, 0xc00008e060, 0x0, 0x20, 0x169b8a0, 0xc000050901, 0xc00008e060)
	/usr/local/go/src/math/big/intconv.go:190 +0x66
math/big.(*Int).setFromScanner(0x0, 0x1887c40, 0xc00008e060, 0x0, 0xc0004bf388, 0x100f0bb)
	/usr/local/go/src/math/big/int.go:431 +0x4d
math/big.(*Int).SetString(0x0, 0xc00009b1f0, 0x4, 0x0, 0xa481888, 0x0)
	/usr/local/go/src/math/big/int.go:425 +0x89
github.com/candid82/joker/core.scanBigInt(0xc00009b1f0, 0x5, 0xc00009b1f0, 0x4, 0x0, 0xc0000c0000, 0xc0004bf4c8, 0x1154131)
	/Users/craig/go/src/github.com/candid82/joker/core/read.go:299 +0x65
github.com/candid82/joker/core.readNumber(0xc0000c0000, 0x7200000031, 0x3e0000003d)
	/Users/craig/go/src/github.com/candid82/joker/core/read.go:397 +0x3e7
github.com/candid82/joker/core.Read(0xc0000c0000, 0xc000000031, 0x105648f, 0x7)
	/Users/craig/go/src/github.com/candid82/joker/core/read.go:1405 +0x165
github.com/candid82/joker/core.TryRead(0xc0000c0000, 0x0, 0x0, 0x0, 0x0)
	/Users/craig/go/src/github.com/candid82/joker/core/read.go:1508 +0xd7
main.processReplCommand(0xc0000c0000, 0x3, 0xc0004bfcd8, 0xc00006bda0, 0x3)
	/Users/craig/go/src/github.com/candid82/joker/main.go:166 +0x45
main.repl(0x3)
	/Users/craig/go/src/github.com/candid82/joker/repl.go:133 +0x80f
main.main()
	/Users/craig/go/src/github.com/candid82/joker/main.go:894 +0x1439

user=>
```

### DIFF
--- a/core/read.go
+++ b/core/read.go
@@ -8,6 +8,7 @@ import (
 	"math/big"
 	"math/rand"
 	"regexp"
+	"runtime/debug"
 	"strconv"
 	"unicode"
 	"unicode/utf8"
@@ -109,6 +110,10 @@ func MakeReadError(reader *Reader, msg string) ReadError {
 		filename: reader.filename,
 		msg:      msg,
 	}
+}
+
+func MakePanicString(r interface{}) string {
+	return fmt.Sprintf("Unexpected panic %q at:\n%s", r.(error), debug.Stack())
 }
 
 func MakeReadObject(reader *Reader, obj Object) Object {
@@ -1491,7 +1496,12 @@ func TryRead(reader *Reader) (obj Object, err error) {
 	defer func() {
 		if r := recover(); r != nil {
 			PROBLEM_COUNT++
-			err = r.(error)
+			switch r.(type) {
+			case ReadError:
+				err = r.(error)
+			default:
+				err = MakeReadError(reader, MakePanicString(r))
+			}
 		}
 	}()
 	for {


candid82:
Hm... So it used to crash on unexpected errors, but for whatever reason I changed it [here](https://github.com/candid82/joker/commit/5da9fb37bb69dd3f5947ae10d3af72ed255e8fa7). I don't remember why. `TryParse` does crash on unexpected errors. I think crashing it the right behavior. Thoughts?

jcburley:
> Hm... So it used to crash on unexpected errors, but for whatever reason I changed it [here](https://github.com/candid82/joker/commit/5da9fb37bb69dd3f5947ae10d3af72ed255e8fa7). I don't remember why. `TryParse` does crash on unexpected errors. I think crashing it the right behavior. Thoughts?

Maybe the idea was that TryRead "wants" to be more lenient and continue reading even in the face of any error (note that it increments PROBLEM_COUNT)? I don't know offhand what use cases might apply, but for formatting, that might make sense, at least for _expected_ errors.

For _unexpected_ errors (those not panic'ed by Joker itself as a means to signal problems with user input), it's helpful to report the stack dump, even if only in a "developer mode".

It's also reasonable to consider the scope of _unexpected_ errors to be much smaller within `TryParse` then elsewhere in Joker, when it comes to assessing impact on internal data structures and such that are necessary for reasonably continuing processing.

I.e. `TryParse` is probably more likely to be able to safely continue processing further input, even after hitting a segfault or something, then `TryRead`, `TryEval`, and so on.

So I don't find myself terribly concerned about its packaging up unexpected errors into an `error` object that it then returns to the caller, because that might be just the ticket for its use cases (outside of when being called in conjunction with reading and especially evaluating) and it'll probably continue to work just fine, as I've observed it when I managed to crash it working on the `math/big` changes (to use pointers).

candid82:
@jcburley I decided to [change](https://github.com/candid82/joker/commit/8ed8c082bb501c55627c2d6f17b9ed916bd24fc3) TryParse to panic on unexpected errors. It should address the issue you ran into (it will print full stacktrace when it crashes). If we see any unintended consequences of this change we can revisit this PR then. Closing for now. Thanks!

PR #462, opened 2021-04-07
Keep on Linting After Parse Errors

jcburley:
The first commit herein, changing the `macro-call` linter test, illustrates what seems to be an inconsistency with regard to whether linting (or, more generally, parsing without evaluation) continues after an error.

This PR tries to clean up some such inconsistencies, but it's not 100% clear to me this is The Right Thing To Do. Still, it might be what people reasonably expect when linting (e.g. that a malformed macro invocation doesn't shut down further linting).

A potential problem with this is "runaway errors", such as when an early parsing error causes a deluge of downstream errors for code that's just fine once that early error is fixed.

The Go compiler mitigates such situations with a "too many errors" diagnostic after some number of errors. That might be worth considering, and I'm happy to implement that, even if what constitutes a "potentially fatal" diagnostic (contributing to such a count and potential aborting of further linting) versus a mere "error" might remain difficult to apprehend.

### DIFF
--- a/core/procs.go
+++ b/core/procs.go
@@ -1782,10 +1782,12 @@ func ProcessReader(reader *Reader, filename string, phase Phase) error {
 		expr, err := TryParse(obj, parseContext)
 		if err != nil {
 			fmt.Fprintln(Stderr, err)
-			return err
 		}
 		if phase == PARSE {
 			continue
+		}
+		if err != nil {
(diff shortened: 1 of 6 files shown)


candid82:
I guess this makes sense. Let's give it a try and see how well it works. Thanks!

Issue #460, opened 2021-04-01
Support Arbitrary BigFloat Precision for stdlib Packages and "M"-Suffixed Constants


jcburley:
Add `MakeBigFloat` API (to `core`) so stdlib packages can specify `^BigFloat` (for definitions of constants); also support arbitrary precision (based on number and widths of mantissa digits) for such constants (ending in `M`), instead of setting precision to a constant 256 as Joker currently does.

This work is already done in https://github.com/jcburley/joker/releases/tag/gostd-v0.10: it computes the precision of a `BigFloat` constant based on the number and “width” of digits in the constant (minimum being `53`, the precision for a `float64` aka `Double`); and a new core function `precision` returns the precision info.

I’m volunteering to crossport this over here and also to create a `set-precision` function that’d work on BigFloat if there’s any interest.

(I think there might be more work that could be done to preserve/extend precision on operations involving `BigFloat`; this seems like an especially useful starting point, at least.)

candid82:
Frankly I doubt many people use `BigFloat`, but having a better support for it would not hurt. The improvements you described make sense to me.

jcburley:
> Frankly I doubt many people use `BigFloat`, but having a better support for it would not hurt. The improvements you described make sense to me.

Okay, I'll work up a PR to bring that support over sometime soon!

PR #463, opened 2021-04-20
Support Arbitrary BigFloat precision and precision-manipulating functions

jcburley:
Fixes https://github.com/candid82/joker/issues/460.

Is `set-precision` the right name? It almost seems more like a `with-precision`, except that doesn't seem right either.

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,7 +102,9 @@ There is [Sublime Text plugin](https://github.com/candid82/sublime-pretty-edn) t
 | Symbol     | n/a                   |
 | Time       | time.Time             |
 
-Note that `Nil` is a type that has one value `nil`.
+See [Floating-point Constants and the BigFloat Type](docs/misc/bigfloat.md) for more on `BigFloat` (`M`-suffixed) constants.
+
+Note that `Nil` is a type that has one value: `nil`.
 
 1. The set of persistent data structures is much smaller:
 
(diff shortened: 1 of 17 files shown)


README.md
```diff
@@ -128,6 +128,70 @@ Note that `Nil` is a type that has one value `nil`.
 - Map entry is represented as a two-element vector.
 - resolving unbound var returns `nil`, not the value `Unbound`. You can still check if the var is bound with `bound?` function.
 
+### Floating-point Constants and the BigFloat Type
```
candid82, line 131: I really appreciate this thorough explanation. I do think this is a fairly niche topic that is probably better placed in a separate file in `docs/misc` directory and linked from here.

core/data/core.joke
```diff
@@ -4538,3 +4538,31 @@
   ([] (exit 0))
   ([^Int code]
    (exit__ code)))
+
+(defn ^BigInt precision
```
candid82, line 4542: It feels like `precision` and `set-precision` belong in `joker.math`. Would you agree?

core/object.go
```diff
@@ -1003,6 +1003,75 @@ func (bi *BigInt) Compare(other Object) int {
...
+		s = s[1:]
+	}
+	if s == "Inf" || s == "inf" || s == "NaN" {
+		return
+	}
+
+	bitsNeeded := 0.
+
+	// Assume base 10 at first.
+	bitsPerDigit := 3.3 // (joker.math/log-2 10) => 3.32192809488736
```
candid82, line 1024: Should not it be more precise value (and rounded up) like 3.322 or otherwise it may not be enough to represent all digits?

core/procs.go
```diff
@@ -1870,6 +1871,40 @@ var procIsNamespaceInitialized = func(args []Object) Object {
 	return MakeBoolean(found && ns.Lazy == nil)
 }
 
+var procPrecision = func(args []Object) Object {
+	prec := int64(-1)
+	switch n := args[0].(type) {
+	case Number:
+		switch n := n.(type) {
```
candid82, line 1878: Why nested switch on type for the same value?

core/procs.go
```diff
@@ -1870,6 +1871,40 @@ var procIsNamespaceInitialized = func(args []Object) Object {
...
+var procSetPrecision = func(args []Object) Object {
+	p := EnsureArgIsNumber(args, 0).Int().I
+	if p < 0 {
+		panic(RT.NewArgTypeError(0, args[0], "non-negative Int"))
+	}
+	prec := uint(p)
+	switch n := args[1].(type) {
+	case Number:
+		switch n := n.(type) {
+		case *BigFloat:
```
candid82, line 1901: The same here, why double switching on type? Perhaps `EnsureArgIsBigFloat` and `ExtractBigFloat` need to be created, similar to helpers for other types?

candid82: Actually, `EnsureArgIsBigFloat` is auto-generated...

README.md
jcburley: Done!

core/data/core.joke
jcburley: I wrestled with this question. I expect anyone using these functions for "real work" would be pulling in `joker.math` anyway, so that isn't a concern here; but, since these functions pertain to the builtin (one might say "core") `BigFloat` type, they seem kinda-core-ish to me, as are `+` and many other math functions that are nevertheless fundamental.

On the other hand, anything Joker introduces to `joker.core` risks someday conflicting with whatever might go into `clojure.core`.

Since I don't have a strong opinion, I'll go with whichever you like. Just let me know!

core/object.go
jcburley: Huh, I just plain rounded up incorrectly. `3.33` seems better. Fixed!

core/procs.go
jcburley: Not sure what i was thinking there; maybe that I might later decide to support non-numeric types (?!). Fixed.

core/procs.go
jcburley: Yup, that's an easy fix.

core/procs.go
jcburley: Note that I also decided to not return `-1`, but rather panic on, any unsupported type. That's more-typical Clojure/Joker behavior.

core/data/core.joke
candid82: My approach is to not put anything in the core namespace unless it's obvious it belongs there. Let's move these two functions into joker.math.

jcburley: > My approach is to not put anything in the core namespace unless it's obvious it belongs there. Let's move these two functions into joker.math.

Sounds good...done, though it was more involved (due to trying to keep/make things clean) than I expected when I plunged into this.

candid82:
Thanks a lot!

PR #464, opened 2021-06-10
Support Go1.17beta1

jcburley:
This is not urgent nor even strictly necessary (yet), but it keeps things clean (so source files don't get changed) when building Joker with go1.17beta1, and also updates normal source files as a result of having run `go fmt ./...` using that version of Go.


candid82:
Thanks!

PR #465, opened 2021-07-07
have read__ wrap io.Reader

youngnh:
today, if you write `(read (os/open f))` then Joker will crash telling you that it expected a RuneReader but that you gave it a File

so this PR implements a very small change to `read` that will wrap `io.Reader` (which File implements) in bufio.NewReader (which implements RuneReader)

### DIFF
--- a/core/procs.go
+++ b/core/procs.go
@@ -1147,7 +1147,7 @@ var procRead = func(args []Object) Object {
 	default:
 		panic(RT.NewArgTypeError(0, args[0], "io.RuneReader or io.Reader"))
 	}
-	return readFromReader(f)
+	return readFromReader(r)
 }
 
 var procReadString = func(args []Object) Object {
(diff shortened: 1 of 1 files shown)


core/procs.go
```diff
@@ -1138,7 +1138,15 @@ func readFromReader(reader io.RuneReader) Object {
...
+	var r io.RuneReader
+	switch f := args[0].(type) {
+	case io.RuneReader:
+		r = f
+	case io.Reader:
+		r = bufio.NewReader(f)
+	default:
+		panic(RT.NewArgTypeError(0, args[0], "io.RuneReader or io.Reader"))
+	}
 	return readFromReader(f)
```
candid82, line 1150: CI fails because `f` is undefined here. Should this be `return readFromReader(r)`?

youngnh: Yup 🤦‍♂️

candid82:
Thanks!

PR #473, opened 2021-10-08
Add runtime lib

jcburley:
Is this possibly useful to anyone else? It could help my processes out a little bit (they currently use some kludges instead), though I'm planning to transition to my `gostd` fork of Joker soon, at which point I can just use its autogenerated `go.std.runtime` namespace directly (so please don't merge this just for my use).


candid82:
I doubt this is useful for regular users (with the possible exception of `joker-version`, but it's covered by `joker.core/joker-version`). On the other hand this namespace (`joker.runtime`) could be used in the future for potentially useful functions. 
Does adding runtime package dependency increase the binary size significantly or is the increase negligible?

candid82:
I guess it doesn't hurt to have it then. Thanks!

PR #474, opened 2021-10-14
Fix eduction definition.

temochka:
`xform*` has no special syntactic meaning in Clojure/joker and thus valid uses like `(eduction (map inc) (map inc) coll)` get flagged for argument count mismatch. Looks like joker correctly adjusts this definition [for ClojureScript](https://github.com/candid82/joker/blob/53f815b06e61a633740bfbb70bc08c40251e391c/core/data/linter_cljs.joke#L46) but not Clojure.

### DIFF
--- a/core/data/linter_clj.joke
+++ b/core/data/linter_clj.joke
@@ -80,7 +80,7 @@
 
 (defn ensure [ref])
 (defn unchecked-remainder-int [x y])
-(defn eduction [xform* coll])
+(defn eduction [& xforms])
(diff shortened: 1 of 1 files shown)


candid82:
Thanks!
RE: `linter.clj`. I was able to fix `eduction` definition in `linter.clj` to suppress the warning. Are you using the latest version of Joker?

temochka:
@candid82 tested on v0.17.1

PR #476, opened 2021-12-23
json-seq: add the keywords option

GSawko:
Make the `json-seq` function accept the `keywords?` option for symmetry with `read-string`.


candid82:
Thanks!

PR #480, opened 2022-07-15
Support explicit environment in joker.os/exec

### LINKED ISSUE #479: large-forked-stdout Fails with Go 1.19rc2
This test attempts to run, within the forked process, the `joker` binary in the top-level directory, depending on `$PWD` to point to that top-level directory.

However, the `run-eval-tests.joke` script calls `joker.os/exec` with `:dir` pointing into the test subdirectory; and `os/exec.Command`, having left the returned `Cmd.Env` field `nil` and with the intervening Joker code not populating it, causes `os/exec.(*Cmd) Start` to substitute the "new" (correct) value for `$PWD`, resulting in a not-found error:

```
input.joke:6:11: Eval error: fork/exec /home/craig/go/src/github.com/candid82/joker/tests/eval/large-forked-stdout/joker: no such file or directory
```

This change in behavior is [documented](https://tip.golang.org/doc/go1.19#:~:text=A%20Cmd%20with%20a%20non%2Dempty%20Dir%20field%20and%20nil%20Env%20now%20implicitly%20sets%20the%20PWD%20environment%20variable%20for%20the%20subprocess%20to%20match%20Dir.) for Go 1.19.rc2. That link will surely change, so here is the current pertinent content:

> [os/exec](https://tip.golang.org/pkg/os/exec/)
> A [Cmd](https://tip.golang.org/pkg/os/exec/#Cmd) with a non-empty Dir field and nil Env now implicitly sets the PWD environment variable for the subprocess to match Dir.

jcburley:
This adds support for a `:env` hash key when calling `joker.os/exec`. When its value is `true`, the current environment (as might be returned by `joker.os/env` as a hash) is used for the new process's environment. When unspecified or its value is `false`, the behavior of Joker itself is unchanged; but the Go runtime might, as of 1.19, set the `PWD` environment variable, in that new environment, to the value of `:dir` (if specified), changing the behavior of some existing code.

Fixes #479 

It might make sense to someday either EOL this capability (ignore `:env`) or extend it to support a hash or even an array (which would be the lower-level array of strings containing `key=value` settings).

Note that this doesn't fix any existing code, using `joker.os/sh` or `joker.os/sh-from`, that depends on the new environment having the old value of `PWD`.

### DIFF
--- a/std/os.joke
+++ b/std/os.joke
@@ -159,7 +159,7 @@
       :out - string capturing stdout of the program,
       :err - string capturing stderr of the program."
   {:added "1.0"
-   :go "sh(\"\", nil, nil, nil, name, arguments)"}
+   :go "sh(\"\", nil, nil, nil, nil, name, arguments)"}
   [^String name & ^String arguments])
 
 (defn sh-from
@@ -171,13 +171,16 @@
       :out - string capturing stdout of the program,
       :err - string capturing stderr of the program."
   {:added "1.0"
-   :go "sh(dir, nil, nil, nil, name, arguments)"}
+   :go "sh(dir, nil, nil, nil, nil, name, arguments)"}
   [^String dir ^String name & ^String arguments])
 
 (defn exec
   "Executes the named program with the given arguments. opts is a map with the following keys (all optional):
-  :args - vector of arguments (all arguments must be strings),
-  :dir - if specified, working directory will be set to this value before executing the program,
+  :args - vector of arguments (all arguments must be strings).
+  :dir - if specified, working directory will be set to this value before executing the program.
+  :env - if specified and true, the environment will be set to the current environment before executing the program. Otherwise,
+    the environment will be determined by the Go runtime which, if :dir is specified, might or might not map $PWD to the expected
+    value -- it might be the current $PWD but, as of Go 1.19, is expected to be set to the value of :dir.
   :stdin - if specified, provides stdin for the program. Can be either a string or an IOReader.
   If it's a string, the string's content will serve as stdin for the program. IOReader can be, for example,
   *in* (in which case Joker's stdin will be redirected to the program's stdin) or the value returned by (joker.os/open).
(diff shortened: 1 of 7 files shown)


jcburley:
Note that I'm not 100% sold on `:env true` as a clean way to specify the current environment for the new process; it just seemed quick and easy.

Alternatives include `:current-env true`, `:env :current`, `:env (joker.os/env)` (supporting converting an arbitrary hashmap to `[]string`), or even having the default be to pass the current environment explicitly (meaning the behavior wouldn't change in Go 1.19) unless `:env nil` is specified.

candid82:
@jcburley good catch! I wonder if adding `:env` argument is worth the trouble though. It seems to me that relying on `$PWD` pointing to top-level directory  inside `large-forked-stdout` test was a bad idea to begin with and it worked almost by accident. Perhaps we should just fix the test?

jcburley:
> @jcburley good catch! I wonder if adding `:env` argument is worth the trouble though. It seems to me that relying on `$PWD` pointing to top-level directory inside `large-forked-stdout` test was a bad idea to begin with and it worked almost by accident. Perhaps we should just fix the test?

"Why not both??".

In the meantime, yes, I think the test can be easily fixed to not rely on this.

Meanwhile, I'm kinda struggling with this vague sense that I've run into this, or a similar problem, before, and perhaps reported it? Yet can't find any record of it while searching. Not sure whether it was in the context of Joker, Golang, or what. But I've often experienced _deja vu_ in my software career, lol....

candid82:
> "Why not both??".

I'd rather not clutter the API with an option that doesn't seem useful outside of addressing an edge case that can be easily avoided. Or do you think this option could find other use cases among "general public"? 

> But I've often experienced deja vu in my software career, lol....

Hear! Hear!

jcburley:
> > "Why not both??".
> 
> I'd rather not clutter the API with an option that doesn't seem useful outside of addressing an edge case that can be easily avoided. Or do you think this option could find other use cases among "general public"?
> 
> > But I've often experienced deja vu in my software career, lol....
> 
> Hear! Hear!

Oh, the `:env` option would certainly be useful beyond just this situation, but it doesn't seem to be needed here, and if desired, it's best to add it properly (supporting at least a hashmap akin to what `joker.os/env` returns, if not also the lower-level array equivalent with embedded `=` separating keys and values, closer to the native value expected by the Go APIs).

PR #481, opened 2022-07-18
Pass path to Joke executable as arg to eval/*/input.joke

jcburley:
Fixes https://github.com/candid82/joker/issues/479 by not relying on `$PWD` being set as previously expected (before Go 1.19), which wasn't quite correct anyway, across a fork/exec.


candid82:
Thanks!

PR #487, opened 2023-07-16
Add control over markdown rendering options

wnh:
I've been building a little static site generator with joker and was missing the ability to turn off the hard line wraps. I've added flags for the other rendering options just for good measure.

I hope the style is ok. I'm not sure what the right pattern is for throwing an exception from the native Go code when a type is wrong. If there is a nicer way I'll happily update it. 

Thanks!

### DIFF
--- a/std/markdown.joke
+++ b/std/markdown.joke
@@ -4,7 +4,14 @@
   markdown)
 
 (defn ^String convert-string
-  "Returns the HTML rendering of Markdown string s"
+  "Returns the HTML rendering of Markdown string s `opts` is an optional map of boolean rendering options (all default to tru to `true`)
+
+  :with-hard-wraps? - Render newlines as <br>.
+  :with-xhtml? - Render as XHTML.
+  :with-unsafe? - When false, all raw html will be omitted from the output. When true html is passed through unchanged.
+  "
   {:added "1.0"
-  :go "convertString(s)"}
-  [^String s])
+   :go {1 "convertString(s)"
+        2 "convertStringOpts(s, opts)"}}
+   ([^String s])
+   ([^String s ^Object opts]))
--- a/std/markdown.joke
+++ b/std/markdown.joke
@@ -4,7 +4,8 @@
   markdown)
 
 (defn ^String convert-string
-  "Returns the HTML rendering of Markdown string s `opts` is an optional map of boolean rendering options (all default to tru to `true`)
+  "Returns the HTML rendering of Markdown string s.
+  opts is an optional map of boolean rendering options (all default to true)
 
   :with-hard-wraps? - Render newlines as <br>.
   :with-xhtml? - Render as XHTML.
--- a/std/markdown.joke
+++ b/std/markdown.joke
@@ -15,4 +15,4 @@
    :go {1 "convertString(s)"
         2 "convertStringOpts(s, opts)"}}
    ([^String s])
-   ([^String s ^Object opts]))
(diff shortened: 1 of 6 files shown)


std/markdown.joke
```diff
@@ -4,7 +4,14 @@
   markdown)
 
 (defn ^String convert-string
-  "Returns the HTML rendering of Markdown string s"
+  "Returns the HTML rendering of Markdown string s `opts` is an optional map of boolean rendering options (all default to tru to `true`)
```
candid82, line 7: Missing `.` after `string s`. Also typo: `to tru`. Finally, there is no markdown support in docstrings, so "`" should not be used. So something like:

> "Returns the HTML rendering of Markdown string s. opts is an optional map of boolean rendering options (all default to true)"

std/markdown.joke
```diff
@@ -4,7 +4,14 @@
...
+  :with-unsafe? - When false, all raw html will be omitted from the output. When true html is passed through unchanged.
+  "
   {:added "1.0"
-  :go "convertString(s)"}
-  [^String s])
\ No newline at end of file
+   :go {1 "convertString(s)"
+        2 "convertStringOpts(s, opts)"}}
+   ([^String s])
+   ([^String s ^Object opts]))
```
candid82, line 17: You can specify `Map` type directly here, which will simplify the implementation: `[^String s ^Map opts]`. See [joker.http/send](https://github.com/candid82/joker/blob/master/std/http.joke#L22) for example.

std/markdown/markdown_native.go
```diff
@@ -33,3 +36,51 @@ func convertString(source string) string {
 	}
 	return buf.String()
 }
+
+func getKeywordFlag(opts Map, name string, def bool) bool {
+	ok, entry := opts.Get(MakeKeyword(name))
+	if !ok {
+		return def
+	}
+	val, ok := entry.(Boolean)
```
candid82, line 45: Please use `EnsureObjectIsBoolean` here for uniformity.

std/markdown/markdown_native.go
```diff
@@ -33,3 +36,51 @@ func convertString(source string) string {
...
+	if flag := getKeywordFlag(options, "with-hard-wraps?", true); flag {
+		renderOptions = append(renderOptions, html.WithHardWraps())
+	}
+	if flag := getKeywordFlag(options, "with-xhtml?", true); flag {
+		renderOptions = append(renderOptions, html.WithXHTML())
+	}
+	if flag := getKeywordFlag(options, "with-unsafe?", true); flag {
+		renderOptions = append(renderOptions, html.WithUnsafe())
+	}
+	md := goldmark.New(
```
candid82, line 68: This is largely a duplication of `convertString`. Can you refactor the code to avoid duplication?

candid82 (commented):
Looks good overall, I made a few suggestions on how to improve some things. Thanks!

std/markdown.joke
wnh: Thanks! I probably should have looked harder for an example of that

std/markdown/markdown_native.go
wnh: When I copied the whole method I definitely told myself I would need to go back and fix the duplication. I guess I just got a little excited 😬

wnh:
Ok... I think I've fixed all those issues. Thanks for your help!

candid82:
@wnh Looks great! Thanks a lot!

PR #490, opened 2023-11-19
Create a_data.go in a reproducible way

bmwiedemann:
Create `a_data.go` in a reproducible way.
We sort to counter golang's random map iteration order.

Note that there is another issue left in the generation of `a_linter_all_data.go`.

This PR was done while working on [reproducible builds for openSUSE](https://en.opensuse.org/openSUSE:Reproducible_Builds).


candid82:
Thanks!

PR #494, opened 2023-12-25
Add day of year to joker.time

romaindoumenc:
A simple function in Go time’s package – but useful if you are using a “year-day” format in your script (e.g. I commonly use format like `2023_344.backup` for timestamp on files).

I did not see tests specific to the time package, but running the script manually looking ok:

```
user=> (joker.time/day-of-year (joker.time/now))
359
```


candid82:
Thanks!

PR #498, opened 2024-02-29
fix empty arrays being serialized as null when writing json string

eko234:


candid82:
Thanks!

eko234:
Np, lovely language btw, I really like your approach about the scope.

PR #506, opened 2024-07-05
Build executables for Free/Net/OpenBSD

dbohdan:
Could you attach BSD executables to future releases? This PR updates `build-all.sh` to build them.

### DIFF
--- a/all-tests.sh
+++ b/all-tests.sh
@@ -2,7 +2,7 @@
 
 fail=0
 
-for test_script in `ls *-tests.sh`
+for test_script in *-tests.sh
 do
     if [ $test_script != "all-tests.sh" ]; then
         echo >&2 "RUNNING $test_script"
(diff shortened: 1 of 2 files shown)


candid82:
@dbohdan Thanks! I'll give it a try.

candid82:
@dbohdan Could you please test BSD binaries from [here](https://github.com/candid82/joker/releases/tag/v1.4.0)?

dbohdan:
You're welcome! Thanks for merging this. It is a convenient change for me. I have [added Joker to my list](http://dbohdan.sdf.org/#joker-official) of projects with official NetBSD binaries.

I can confirm the BSD binaries work on their platforms. The OpenBSD binary works on OpenBSD 7.3 but not 7.5 because the ABI has changed. OpenBSD provides no stable ABI or ABI compatibility shims, so the binaries are more short-lived than those for FreeBSD and NetBSD. A newer version of Go will build for 7.5 (see https://github.com/golang/go/issues/63900).

